PR #256, opened 2016-02-22
Fixes #232

### LINKED ISSUE #232: Deleting a log should return to cache page with proper URL
If I delete a log entry on a cache page, I am returned to the cache description page. (as in what it is shown on screen) 
However, the URL is always http://www.opencaching.nl/removelog.php
The URL should be that of the cache. 

When using the "edit" function, after editing and submitting changes, the user is returned to the cache page and the proper URL is shown in location bar.

mzylowski:


mzylowski:
Any feedback about this?
@andrixnet Your problem is solved by this commit?

kojoty:
looks good, thanks Czarodziej 

PS . if you don't know Czarodziej is an oc.pl nick of @mzylowski

andrixnet:
Yes, I can confirm. Deleting a log returns to viewcache.php correctly.
Thank you.

harrieklomp:
This delete a log is still only possible when Polish or English language is selected. Other languages result in a error page.

PR #266, opened 2016-02-27
Event icons in cloud on map

mzylowski:
Litte commit about issue from #15 
There is a lot work to do but icons are ready now.

Before:
zaznaczenie_005
After: 
zaznaczenie_006


kojoty:
good job!, Thanks @mzylowski

PR #271, opened 2016-02-29
New config for uploading picture sizes

harrieklomp:
Via settingsDefault.inc.php it is possible to set the size for uploaded pictures. Also there are more things to setup (look at bottom of file settingsDefault.inc.php).
It is tested on OC.NL and there we can upload pictures up to 10MB.

Please check if all is correct and change if necessary.


kojoty:
Looks great generally :) 

A few fast observations / comments:
- do we need also change in lib/languages/pl.php?
- we need to add proper config objects in our production config files (settings.inc.php) at all three nodes (pl/nl/ro) before merge this commits - am I right?
- what do you (all) think about accessing to the config through _lib/Objects/OcConfig_ - I think it could be a good idea to use it in new commits and step-by-step remove old config access style...? 

...and the general postulate: I think it would be great to not deliver separate commits with the same comment - then history of files will be more clear and useful - my proposition is to unite commits in such works - with git it seems to be quite easy (there is for example -amend option for commit or it is possible to merge many commits from one branch to another with --no-commit and then commit all changes as one commit etc...) - what do you think about it?

harrieklomp:
A view weeks ago Andrei (oc.ro) did make these changes on oc.nl for me so i could test it.

- lib/languages/pl.php. Line 854 (pictures_intro) and 855 (pictures_intro2) could be translated like in English file.
- Not all developers have access to Polish settings.inc.php, therefor settingsDefault.inc.php does override this.
- I do not know how this lib/Objects/OcConfig works so i cannot say if this should be a good idea.

I am glad that i manage this pull request, this was the first time :-)

kojoty:
Right, settings.inc.php overrides settingsDefault, so everything should work.
I'll make a change in pl translation file after your commit 
we will come back to the subject of lib/Objects/OcConfig in a future I hope :)

**my opinion: go ahead and merge it! Thanks!**

harrieklomp:
That will be done tomorrow. Now at work untill 6 AM.

Issue #214, opened 2015-11-02
Automatyczne tłumaczenia opisów nie działają


azhag-pl:
Jeśli przełączymy język strony na inny niż polski, to zamiast opisu strony jest tylko:

> Automatic translation thanks to:
> Technologia Google™

I nic poza tym. Jeśli Google pokręciło i nie da się z tym nic zrobić, to już lepiej zostawić oryginalny opis.

kojoty:
It seems that google introduced paid access to translation api and closed the free version:

> What happened to earlier free versions of Translate API?
> Google Translate API v1 is no longer available as of December 1, 2011 and has been replaced by Google Translate API v2. Google Translate API v1 was officially deprecated on May 26, 2011. The decision to deprecate the API and replace it with the paid service was made due to the substantial economic burden caused by extensive abuse.
[https://cloud.google.com/translate/v2/faq#pricing]

This seems also that the only free option to have translation is to use google translation plugin:
[https://support.google.com/translate/answer/2534601?hl=en]

We need to decide if we want to use it...

PR #278, opened 2016-03-02
added the option to disable google translations

following5:
I have set up a (devel) site and would like to see the original cache descriptions here (site lang is English, example descs are German).

### DIFF
--- a/viewcache.php
+++ b/viewcache.php
@@ -1114,14 +1114,14 @@ if ($error == false) {
             $desclang = $_REQUEST['desclang'];
         }
 
-        $enable_google_translation = false;
-
         //is no description available in the wished language?
         if (array_search($desclang, $desclangs) === false) {
             $desclang = $desclangs[0];
         }
 
-        if (strtolower($desclang) != $lang && $lang != 'pl')
+        if (isset($disable_google_translation) && $disable_google_translation)
+            $enable_google_translation = false;
+        else if (strtolower($desclang) != $lang && $lang != 'pl')
             $enable_google_translation = true;
         else
             $enable_google_translation = false;
--- a/viewcache.php
+++ b/viewcache.php
@@ -1119,12 +1119,15 @@ if ($error == false) {
             $desclang = $desclangs[0];
         }
 
-        if (isset($disable_google_translation) && $disable_google_translation)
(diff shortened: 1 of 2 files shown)


viewcache.php
```diff
@@ -1114,14 +1114,14 @@ function onTheList($theArray, $item)
...
-
         //is no description available in the wished language?
         if (array_search($desclang, $desclangs) === false) {
             $desclang = $desclangs[0];
         }
 
-        if (strtolower($desclang) != $lang && $lang != 'pl')
+        if (isset($disable_google_translation) && $disable_google_translation)
+            $enable_google_translation = false;
+        else if (strtolower($desclang) != $lang && $lang != 'pl')
```
wloczynutka: Could you check why 'pl' is hardcoded? i suppouse this is a bug. Maybe it is possible to fix it?

viewcache.php
```diff
@@ -1114,14 +1114,14 @@ function onTheList($theArray, $item)
...
 
-        $enable_google_translation = false;
-
         //is no description available in the wished language?
         if (array_search($desclang, $desclangs) === false) {
             $desclang = $desclangs[0];
         }
 
-        if (strtolower($desclang) != $lang && $lang != 'pl')
+        if (isset($disable_google_translation) && $disable_google_translation)
```
wloczynutka: if / else should be always used witch braces { } (PSR-2 http://www.php-fig.org/psr/psr-2/)

viewcache.php
following5: I have no idea why _$lang != pl_ was added. Effectively, it disables automatic translation if the user choses Polish language. Does not look like a bug but some poor coding.

viewcache.php
following5: The original code was without braces. I have adjusted it to PSR-2 now.

wloczynutka:
Hello everyone. I'm merging this PR, but please could you check if automatic google translation is really working anywhere? I remember that Google tranmsaltion was not working some time ago, but later I don't use it so I don't know.

kojoty:
translations are not working - for reference the issue is here  #214

andrixnet:
I think I also discussed at some point the fact that automated translations with Google Translate do not work.

PR #280, opened 2016-03-02
fix admin menu report statistics on newlogs and logmap page

following5:
With commit 02d2f6b, a `db_disconnect()` was inserted at the top of newlogs.php. As a result, the mysql_ calls at line 179ff of main.tpl do not work. They produce PHP warnings (if enabled) when evaluating the template code, and the numbers in the admin menu are missing on the new logs page.

There is no other page with does a `db_disconnect()`, except for index.php which does it at the very end. So I think for consistency, this problem can be solved by removing this statement.

### DIFF
--- a/newlogs.php
+++ b/newlogs.php
@@ -7,8 +7,6 @@ if (!isset($rootpath))
 //include template handling
 require_once ($rootpath . 'lib/common.inc.php');
 require_once($stylepath . '/lib/icons.inc.php');
-// this file use only database class. We don't need mysql_ connectors anymore.
-db_disconnect();
 
 //Preprocessing
 if ($error == false) {
--- a/logmap.php
+++ b/logmap.php
@@ -2,7 +2,6 @@
 
 global $dateFormat;
 require_once ('./lib/common.inc.php');
-db_disconnect();
 
 //Preprocessing
 if ($error == false) {


following5:
Oops, same problem in logmap.php. I have added it to this branch. But that's all.

newlogs.php
```diff
@@ -7,8 +7,6 @@
 //include template handling
 require_once ($rootpath . 'lib/common.inc.php');
 require_once($stylepath . '/lib/icons.inc.php');
-// this file use only database class. We don't need mysql_ connectors anymore.
```
wloczynutka: I'm not pretty sure if we should remove this lines. In comment there is explanation why db_disconnect() is placed here. Probably some performance issue. Could you check once again if it generate any notice? maybe there is other way to prevent it?

following5: It always generates notices when newlogs.php or logmap.php is run:

```
Warning: mysql_result() expects parameter 1 to be resource, boolean given in /mnt/ocpl/lib/common.inc.php(611) : eval()'d code on line 179

Warning: mysql_result() expects parameter 1 to be resource, boolean given in /mnt/ocpl/lib/common.inc.php(611) : eval()'d code on line 181

Warning: mysql_result() expects parameter 1 to be resource, boolean given in /mnt/ocpl/lib/common.inc.php(611) : eval()'d code on line 183

Warning: mysql_result() expects parameter 1 to be resource, boolean given in /mnt/ocpl/lib/common.inc.php(611) : eval()'d code on line 185
```

The developer who ported  newlogs and logmap from mysql_ to PDO just went too far by closing the mysql_ connection. He was not aware that it is still needed by the code in main.tpl, line 178-185. The only benefit of closing the mysql_ connection is that it frees a some KB of memory; there is no impact on performance.

> maybe there is other way to prevent it?

Porting lines 178 to 185 of main.tpl to PDO would prevent it, but that would create an additional database connection for all pages of the website, which is waste of resources as long as the mysql_ connection is avaiable.

(Thanks for merging the other commits!)

wloczynutka: Thank you for investigation. Plese check commit https://github.com/opencaching/opencaching-pl/commit/f5257633e3b57962245d552def8e225948fc4089 and main.tpl.php

I removed using mysql functions, this should fix notices.

following5:
Ok, problem  is solved.

PR #281, opened 2016-03-02
get rid of some PHP notices

following5:
There were some PHP notices and warnings when running OCPL code with PHP setting `error_reporting = E_ALL & ~E_DEPRECATED & ~E_STRICT`.

The `session_start()` in cachemap2.php and lib/marker.php is redundant to `session_start()` in common.inc.php.

### DIFF
--- a/cachemap2.php
+++ b/cachemap2.php
@@ -161,15 +161,17 @@ global $lang;
 
 $userid = '';
 
-$get_userid = strip_tags($_REQUEST['userid']);
+if (isset($_REQUEST['userid'])) {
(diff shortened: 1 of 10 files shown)


following5:
I have systematically scanned the site for PHP warnings and notices and think that I spotted most of them. After merging this and #280, it should be fine to enable PHP notices on all OCPL developer installations, and then fix those that I missed. This would help to detect unitialized variables and to improve coding style.

kojoty:
great job! Thanks.

PR #293, opened 2016-03-04
removed obsolete javascript code from cachemap2.php

following5:
The only effect of this code is that it produces a Javascript error when leaving the cachemap2 page:

`TypeError: document.myForm is undefined`

There is no form "myForm" and no form element "time"; must be some leftover code.

### DIFF
--- a/cachemap2.php
+++ b/cachemap2.php
@@ -246,12 +246,6 @@ if ($usr == false) {
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


kojoty:
yeah! map_v2 is obsolete at all and we would like to remove it but...  we have a fans of map_v2 in community.

PR #297, opened 2016-03-04
get rid of some php notices

following5:


kojoty:
great! Thanks for that :)

following5:
I am finished with testing OCPL code, so to avoid introducing new glitches, I would recommend that OCPL developers now enable PHP notices. Depending on the setup, this is done either in php.ini  `error_reporting` or in httpd.conf `php_admin_value error_reporting`.

wrygiel:
I would recommend that too. I can make it a default in the next VM version.

PR #320, opened 2016-03-07
non-working translations based on google removed (in context of #214)

kojoty:
Google translation in version 1 (used in oc code) is not available anymore.
New version (v2) is now a paid service so is useless for our project.

I save a stub of config in the code - maybe we'll decide to use google translation plugin instead of previous translation solution - then it would be useful.


mzylowski:
+1 :>

PR #324, opened 2016-03-09
Db cleanup - removing of unused tables

kojoty:
this is the first part of changes.
It should be safe - tables which I would like to drop are not used by the OC code (or I just remove dead code related to them).


andrixnet:
Database updated on OCRO

PR #325, opened 2016-03-09
Removing of foreign_caches db table and related code

kojoty:
about table foreign_caches:
- it is used to display foreign caches on map_v2 ONLY (foreign means not from current oc node)
- currently there is no mechanizm (code) which can synchronize this data
- data stored here is obsolete - not refreshed for many years


harrieklomp:
This was done via crontab.
Now we have Flopp's maps to show all foreign caches :-)

andrixnet:
Database updated on OCRO

PR #326, opened 2016-03-09
Do not touch images smaller then 100kb #275

mzylowski:
When file is smaller then 100kb we accept them and just save on the server.

It was difficult to test it on VM. I think we need additional test it before merge.

### DIFF
--- a/editpic.php
+++ b/editpic.php
@@ -76,7 +76,7 @@ if ($error == false) {
                         }
 
 
-			if ($config['limits']['image']['resize'] == 1) {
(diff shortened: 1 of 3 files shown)


kojoty:
@mzylowski, it seems that this change should be also done in files: editpic and newpic-gps..
- just locate places where SimpleImage is used to resize

mzylowski:
Oh, sorry. I think in the futher we need one interface for putting picure on the server. :)

andrixnet:
The same changes should also be applied here: 
newpic-gps.php
editpic.php

andrixnet:
@mzylowski your suggestion is good. Please create an issue about it.

PR #333, opened 2016-03-10
Improvement of balloon on v3 map 

mzylowski:
selection_001
My propose for #270

### DIFF
--- a/lib/xmlmap.php
+++ b/lib/xmlmap.php
@@ -77,7 +77,7 @@ class tmp_Xmlmap
         $fields = 'code|name|location|type|size2|' .
                   'recommendations|rating_votes|rating|' .
                   'willattends|status|owner|founds|' .
-                  'notfounds|internal_id';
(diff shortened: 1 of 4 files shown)


lib/xmlmap.php
```diff
@@ -365,6 +365,7 @@ private function htmlFormat(array $params)
 
         tpl_set_var('user_name', $geoCache->getOwner()->getUserName());
         tpl_set_var('user_profile', $geoCache->getOwner()->getProfileUrl());
+        tpl_set_var('start_date', $geoCache->getDatePlaced()->format('d M Y'));
```
wloczynutka: Date format better is take from config class:

OcConfig->getDateFormat()

(Benelux uses diffrent than PL)

mzylowski:
@wloczynutka: thanks ;]
There is a lot of small variables and tips.. Sometimes is very difficult to find all of them.

kojoty:
good job! Thanks!

PR #339, opened 2016-03-13
Oc config renames to PSR-2

kojoty:
this is not necessary but I couldn't hold back


wloczynutka:
:+1:

andrixnet:
Does this require each node to change their settings.inc.php at this time?

kojoty:
@andrixnet, don't worry, any changes in configuration are not necessary at this point

PR #340, opened 2016-03-13
db class - refactoring around #276

kojoty:
This is a 'preview' of way I would like to rewrite the DB access class. 

this is a=only a draft - it wasn't finished nor tested. If you accept this direction I'll finish the works and then commit changes.

My vision in short words:
 - DB classes are moved to `/Utlis/Database` dir (I would like to share with you my vision of code structure - please take a look here:[https://docs.google.com/document/d/1UH0tJ3GXRsDqgPZf8eMsYbE--cVpKPlYuwwK-1yTz6A/])

- `OcPdo` class extends `PDO` and contains:
    - db class initialization specific for OC
    - error handling + debug messages handling code
    - specific debug overrides to PDO (which prints debug messages on calling PDO standard methods) 

- `OcDb` class extends `OcPdo` and contains:
    - custom methods for db accessing (with pdoStmt hidden inside a class code etc.)

**All comments are welcome!**

### DIFF
--- a/Utils/Database/OcDb.php
+++ b/Utils/Database/OcDb.php
@@ -1,0 +1,333 @@
+<?php
+
+class OcDb extends OcPdo
+{
+
+    protected $stmt; //internal PDOStatement TODO
+
+
+    /**
+     * @return one row from result, or FALSE if there are no more rows available
+     * The data is returned as an array indexed by column name, as returned in your
+     * SQL SELECT
+     */
+    public function dbResultFetch()
+    {
+        return $this->stmt->fetch();
+    }
+
+    /**
+     * for queries witch LIMIT 1 return only one row and reset database class preparing it for next job.
+     */
+    public function dbResultFetchOneRowOnly()
+    {
+        $result = $this->stmt->fetch();
+        $this->reset();
+        return $result;
+    }
+
+    /**
+     * @return all rows from result as complex array.
+     * The returned array contains all of the remaining rows (if you have previously called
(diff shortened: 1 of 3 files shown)


wloczynutka:
Conception in doc is very good. This is thing I always missing in oc code. Very good direction.

Database: 
moving to utils/Database OK,

I'm not pretty sure if we really need two database classes. 
Is it possible to having everything in one class?

As I undestand, until classes have public constructors, there is possible creating multiple instances. And - I guess - each instance create own database connection. Some time ago we had problem with too many DB connections. (and this problem repeat from time to time). Maybe database class should be only one and strict limited to singleton, to be sure only one database connection per thread is established?

Or maybe we can limit db connection in other way?
I have no strong knowdlege about this.

kojoty:
> I'm not pretty sure if we really need two database classes. 
> Is it possible to having everything in one class?

I'm sure we can - but the question is why?
Maintenance of the very larg (long) file is harder than two separated, especially if the logic behind the separation is simple and clear.
The next thing is that maybe using of pure OcPdo class can be useful in some situations and then this is possible...

> As I undestand, until classes have public constructors, there is possible creating multiple instances. And - I guess - each instance create own database connection. Some time ago we had problem with too many DB connections. (and this problem repeat from time to time). Maybe database class should be only one and strict limited to singleton, to be sure only one database connection per thread is established?

That's right. Generally I read about disadvantages of singletons, but I think at this stage of works it could be a good idea to prevent creation of the objects in a other ways. I will implement it - we will be able to change it in a future....

PR #342, opened 2016-03-14
added comment on editing log entries

following5:
Added some comment in editlog.php to prevent breaking OKAPI replication of recommendation flags in the future.


following5:
This is the only spot in OCPL code which is critical for OKAPI replication, therefore I think it should be annotated.

PR #361, opened 2016-03-18
Few small fixes

mzylowski:
Any suggestions? ;]


mzylowski:
There are no comments, so merged ;]

lib/languages/pl.php
```diff
@@ -428,7 +428,7 @@
     'newest_caches' => 'Najnowsze skrzynki',
     'total_of_active_caches' => 'Liczba aktywnych skrzynek',
     'total_of_caches' => 'Liczba skrzynek',
-    'total_moved' => 'Liczba przeniesionych skrzynek',
+    'total_moved' => 'Liczba transferów skrzynek mobilnych',
```
kojoty: @mzylowski Have you checked how it sounds in other languages - maybe the translations should be also fixed?

mzylowski: It looked like polish translation problem, but you are right.
In english it is: 'total_moved' => 'Number of moved caches',

kojoty: zastanawiam się też, czy lepiej nie brzmiało by: "liczba przeniesień skrzynek mobilnych"

PR #362, opened 2016-03-19
geolocationn on the map fix

kojoty:
Geolocation is working properly again :)

### DIFF
--- a/lib/cachemap3.js
+++ b/lib/cachemap3.js
@@ -325,7 +325,7 @@ function initCustomControlls( map, params ){
 
                 //geolocation present in browser
                 button.css('display','block'); //show position button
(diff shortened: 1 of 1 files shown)


andrixnet:
what is the darker tint the map (google map at least) gets beyond zoom level 7 or 8?

kojoty:
@andrixnet, I guess you mean those darker (but transparent) square tiles which appears one-by-one on the map when you change the zoom? - if yes those are tiles generated by OKAPI - okapi generates transparent images with caches marks which overlap the main layer of the map.

wrygiel:
This was requested by some users in the early stages of map design. Supposedly, they weren't able to see the cache icons properly, so I needed to make the background darker, for the icons to stand out more.

PR #364, opened 2016-03-19
Better translation for #361, #353

mzylowski:


lib/languages/en.php
```diff
@@ -428,7 +428,7 @@
     'newest_caches' => 'The latest caches',
     'total_of_active_caches' => 'Overall number of active caches',
     'total_of_caches' => 'Overall number of caches',
-    'total_moved' => 'Number of moved caches',
+    'total_moved' => 'Number of transferred mobile caches',
```
kojoty: I don't know, but maybe: "Number of movings of mobile caches"

kojoty:
@harrieklomp, @andrixnet please tak a look at this minor issue: this stat. in user profile shows _"number of transfers"_ (even of the same cache) instead of _"number of mobile caches transferred"_ (where many transfers of cache are counted as one). Maybe it should be also changed in your languages...

harrieklomp:
The Dutch translation is fine for me :-)

lib/languages/en.php
wrygiel: Number of mobile caches moved

lib/languages/pl.php
```diff
@@ -428,7 +428,7 @@
     'newest_caches' => 'Najnowsze skrzynki',
     'total_of_active_caches' => 'Liczba aktywnych skrzynek',
     'total_of_caches' => 'Liczba skrzynek',
-    'total_moved' => 'Liczba transferów skrzynek mobilnych',
+    'total_moved' => 'Liczba przeniesień skrzynek mobilnych',
```
wrygiel: Liczba przeniesionych skrzynek mobilnych

wrygiel: Chociaż w sumie to nie jest to samo... bo jedną skrzynkę można przenieść parę razy. :-/ Ale tak jak to jest teraz, brzmi dość karkołomnie...

wrygiel: Trochę lipa, taka zabawa w słowotwórstwo ;)

lib/languages/en.php
wrygiel: Total number of mobile cache transfers?

andrixnet:
I am still not sure I understand the context properly. 

What does "moved" refer to? 
* Moving a cache as in changing it's coordinates? (any cache)
* Moving a "moving/mobile cache" (specifically)
* other?

lib/languages/en.php
andrixnet: "transfer" in english may suggest here transfering from one user to another, aka adoption.

harrieklomp:
It refers to a user who has moved X caches. When logging a moving cache as "moved"
selectie_144
When i look at oc.us i see...
selectie_146
Than i wonder why we should translate it a different way than the inventors of the English language ;-)

andrixnet:
Ah, now I understand. 
it is about "number" of moving caches that were moved.

Then yes, the original string sounds better.

lib/languages/en.php
andrixnet: Given the clarification by @harrieklomp I believe the original english string is more appropriate in the context.

andrixnet:
Please just copy the english string to RO at this time, I am working on improving the entire english string set and then will solve the romanian set as well. Thank you.

Issue #291, opened 2016-03-04
Option to configure default search (top right)


andrixnet:
I would like to request an option to define the default search type for the quick search box (top right of all pages). 

Currently the default is to search by waypoint. 
My personal experience is that searching for cache name is much more common and would like to set it as default on OCRO.

kojoty:
hmm... I much more frequently use the search by waypoint function... but maybe you're right. 
I think we can check in the stats what is more often chosen...

andrixnet:
maybe on OCPL  it's waypoint.
maybe for other nodes may be different

wloczynutka:
it is not very hard to make automatic recognize if user input was a geocache waypoint or other string (cache name)

changing of script logic depend on node is too much, we should do not go this way. We already heve too much configurable things.

andrixnet:
while keeping the "find by owner / finder" functionality, can it be done to search by waypoint or cache_title by default, without the need to change which radiobox is checked?

wloczynutka:
Yes, this is not so hard. Join search by waypoint witch search by name, and automatic recognize if input were waypoint or name.

andrixnet:
one good feature this search box has so far is the fact that one may search for a waypoint code without the explicit prefix. (like OP12345 can be searched by entering 1234 for example). 
Can this behaviour be retained?

wloczynutka:
rather nope

andrixnet:
gee, that was really, really, useful

wloczynutka:
@andrixnet, you are only one person who complain about this search box, We have few thousand users and noone complain. :-)

andrixnet:
I see.

wrygiel2:
I would agree with this complaint too. This is one of many not-so-well
designed features of OC.

kojoty:
Redesign of the search box would be a nice feature. It's not the most important right now - I agree with the priority -  but let's come back to this in a future - This issue should stay on the list :)

deg-pl:
New look ;)
![search](https://cloud.githubusercontent.com/assets/13884863/13931726/eafa2e8e-efa4-11e5-94f3-c09fffe4e322.png)

deg-pl:
@andrixnet Will "Nume/Cod" be OK?

andrixnet:
@deg-pl yes, it should be ok. 
Thanks. 

Sorry about my late reply, I worked all day on improving the english strings. Still a lot of work to be done and clarifications, but it's getting better.

[...]

PR #375, opened 2016-03-21
Solution for #291

deg-pl:
In searching box is one option "name/waypoint" instead of two: "cache name" and "waypoint"
If you search 1234, you will find:
- cache "Test cache 1234" and cache with OP1234
If you search GC12345, you will find:
- cache with waypoint set to GC12345 and cache "Cache better than GC12345"

Check it please!


deg-pl:
@harrieklomp , @andrixnet could you translate "waypointname_label"?

harrieklomp:
Dutch and German is translated and committed.

PR #391, opened 2016-03-25
synchronized en, pl, nl, ro and de translation files

following5:
* exactly synchronzied the files line by line
* removed some outdated and duplicate entries
* marked all missing entries as comments

I think the comment-style marks for missing entries make them more readable and easier to fill in than in than the old `'' => 'id'` style. For the case that you use tools to proccess these files that are broken by the comments and cannot be adjusted, I can shuffle it back to the old style.


following5:
There are three entries missing in the Romanian translation which look to be obsolete - I could not find them anywhere in the code:

* pt_show_on_main_map
* WeekCache
* list_of_titled_caches_is_empty

To keep everything consistent, I have added them as commented lines to the Romanian file, too.

wloczynutka:
I thik it is better to remove empty transaltion row insted keep it as empty string.
Why?
Because if our translate function not find proper translation identifier, display simle warring ( something like that: "no transaltion with identifier 'indentifierName'" - users can report it, and programmer cam easly find place of the missing transaltion.

If we put in translation files

> 'indentifierName' => ''

then translation function will not display warring message, just empty sting will be displayed.

Or

Keep empty strings as you prosposed, but add to transalting functtions some code reporting missing transaltion (email to developers / system admin)

following5:
> I thik it is better to remove empty transaltion row insted keep it as empty string.

All empty (missing) translation rows in this branch are commented-out (#). So the  translation function will not see them and output warnings.

wloczynutka:
Ah, right.
Sorry, I missed that.

2016-03-25 13:58 GMT+01:00 following5 <notifications@github.com>:

> I thik it is better to remove empty transaltion row insted keep it as
> empty string.
>
> All empty (missing) translation rows in this branch are commented-out (#).
> So the translation function will not see them and output warnings.
>
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub
> <https://github.com/opencaching/opencaching-pl/pull/391#issuecomment-201268807>
>



-- 
__________________________________
*Andrzej 'Łza' Woźniak*

*(+48) 798 835 592*
skype praca: *andrzej.wozniak.traveltech*
skype dom: *wloczynutka*

andrixnet:
I am also working on translations, the english strings. 
Follow my progress here: https://github.com/andrixnet/opencaching-pl/tree/translation. 
Trying to complete the entire english string set before requesting a pull. 

Then I will complete the romanian translation (some of which is still quite incomplete because of the english strings not being understood well enough).

andrixnet:
I believe it to be good practice for all developers when adding a new string, to add it commented out to the other languages, or not commented, as the english variant. They are easier to track down in a diff, rather then resyncing a language file to get missing strings.

following5:
> I believe it to be good practice for all developers when adding a new string, to add it commented out to the other languages,

Yep. This change does exactly this for all cases where developers forgot it. 

> or not commented, as the english variant.

As far as I understood wloczynutka, this should _not_ be done. If it is wanted to output English strings where translations are missing, that would be better done in he program code. Keeping English strings redundant in translation files is no good idea, the redundancies can get out of sync if strings in en.php are changed.

> I am also working on translations, the english strings.
> Trying to complete the entire english string set before requesting a pull. 

Ok. Then I suggest that you complete that first, and then I can rebase this branch on yours. And when this all is merged, the missing translations - which then are all marked by comments - can be completed.

following5:
There will be too many conflicts with https://github.com/andrixnet/opencaching-pl/tree/translation, as changes that I did here are now duplicated there, but partially in other context (especially in de.php).

As noone seems to be interested in this change, I just discard it.

PR #401, opened 2016-03-25
Added brand resources.

andrixnet:
* In his hands (Font)
* Opencaching logo, vector format

Note: the logo has been reworked a bit, after having difficulties
using it for CAM post processing, so think of this as an optimized
version of the vector logo.


andrixnet:
When I first tried to build upon the Opencaching brand, I had difficulties finding and using suitable files. 
After some digging, I finally got some vector files, but because of the way they were composed, several shops sent me back to the drawing board because they considered them unsuitable for CAM purposes (like cutter plotter or laser engraving).
(trying to make some promotional items).

These files are the result of the rework. I hope they can be useful.

Issue #359, opened 2016-03-18
Problems with waypoint downloads


andrixnet:
1) Download in XML format yields opening the XML file in the browser. 
Needs proper HTTP headers to force downloading.

2) XML format contains polish strings for things like cache size, cache type (although selected language is not polish)

3) XML format contains empty description

4) Ozi WPT format has improper HTTP headers. Browsers downloads a file named ".wpt" (notice lack of name part, only extension)

5) WPT format should replace all UTF8 regional characters to latin1 english (or ANSI)

6) WPT format description could be improved by showing OR12345 / D:1.0 / T:1.5 / Size: Micro instead of OR12345/1.0/1.5/Micro

7) KML format references icon URLS different from those on the website
OC website shows icon from http://www.opencaching.ro/tpl/stdstyle/images/cache/unknown-s.png
KML references icon from http://www.opencaching.ro/images/ge/unknown.png
Current icons in KML are not proper images for Google Earth app. See attached screenshot.

8) Not all formats contain the additional waypoints. This is a list of formats not containing additional waypoints: 
TXT, WPT, KML, UAM, OVL, OV2, LOC. 
While some may not support the additional waypoints, they whould be present in at least: 
TXT, WPT, KML

andrixnet:
https://s22.postimg.org/vnxo7eu5d/kml-icon.png

deg-pl:
@andrixnet - I've fixed problem 1 and 2. Check it please.
A'propos problem 3: into tag "desc" is NOT FULL but SHORT decription of the cache. I suppose, that it is correct.

andrixnet:
Why only short description and not both? (short and full)

deg-pl:
As one tag - "desc"?
Which program uses waypoints exported as XML?

andrixnet:
To be honest, I do not know which program. 
I know the option exists to export as XML, thus I tried it out.

deg-pl:
So we could add i.e. tag "longdesc"/"fulldesc" with full description, but full description may contains HTML unfortunately :(

andrixnet:
yes, it could. and should be escaped.

deg-pl:
I'm just testing task list ;)

- [ ] 1. XML - Proper HTML header
- [ ] 2. XML - i18n
- [ ] 3. XML - long description
- [ ] 4. WPT - filename
- [ ] 5. WPT - UTF-8 -> ASCII
- [ ] 6. WPT - format corrections
[...]

PR #405, opened 2016-03-26
Issue #359, points 3, 9, 10, 11, 12

deg-pl:


lib/search.xml.inc.php
```diff
@@ -4,7 +4,7 @@
 
 $encoding = 'UTF-8';
 $distance_unit = 'km';
-$xmlLine = "    <cache>
+$xmlLine = chr(239) . chr(187) . chr(191) ."    <cache>
```
andrixnet: This was noted for TXT format, not XML. It may even be wrong here.

andrixnet:
Confirmed all except point 11.

lib/search.xml.inc.php
deg-pl: Yes, of course.
Corrected -> commit 348de782394791687e4355e42a651fea859e31f1

PR #448, opened 2016-04-01
searchplugin is not used and obsolate code - lets try to remove it

kojoty:
It seems that searchplugin is never used now...

I'm not absolutely sure, so I can't see any other way to discover if anyone needs it than to remove these files and check if anyone fell the lack of them ans ask for revert...

@wloczynutka, @deg-pl, @mzylowski what do you think?


mzylowski:
+1
You are right, this code look like out of date.
I think we can safety delete this files but we can merge it after this weekend.

deg-pl:
Could you check apache access logs? Maybe this code is called by some remote sites/apps?

kojoty:
@deg-pl, 
I checked awstats and I saw only single access per day.
Bow I checked logs at server...
It seems that there are lot of bots access + geokrety.org link to Oc sites by 
`http://www.opencaching.pl/searchplugin.php?userinput=OP8DGH`

From my point of view we can add rewrite rule in .htaccess which translate all links 
`/searchplugin.php?userinput=<oc-waypoint>` 
to
`/viewcache.php?wp=<oc-waypoint>`

Searchplugin has more options of search but I can't see any usage of it.
What do you think?

deg-pl:
I think you're right! :)
+1

kojoty:
Does anyone of you know what do add to htaccess to redirect all requests from
`/searchplugin.php?userinput=<oc-waypoint> `
to
`/viewcache.php?wp=<oc-waypoint>`

I'm not sure how to deal with it...

andrixnet:
Try something like this: 
```
# /searchplugin.php?userinput=<oc-waypoint> ==> /viewcache.php?wp=<oc-waypoint>
RewriteCond %{REQUEST_URI} "^/searchplugin.php?userinput=([A-Z]{2}[0-9A-Z]{4})$" [NC]
RewriteRule ^/searchplugin.php?userinput=([A-Z]{2}[0-9A-Z]{4})$?$ /viewcache.php?wp=$1 [R=301,L,NC]
```

following5:
The searchplugin is a web browser plugin (I think only for Firefox) which allows to search for Opencaching content through the browser's 'search' field. This is the Opencaching.DE version:

following5:
Outdated link. This is the current one: 
http://www.opencaching.de/resource2/misc/mozilla/OpencachingDE.src

kojoty:
@mzylowski have you tested your solution - I tried it without a success...

@following5, I know, but:
- do you know how to install it - I tried a few times and I can't install it to current firefox
- firefox is only one of many browsers now
- it seems you can only search by cache name
- our logs shows that noone uses it (besides links form site geokrety.org, which we will support)

...so in my opinion it is useless now, but this is only my private opinion :)
Do you know if anyone uses it for searching at opencaching.de?

following5:
I neither know how to install it nor if it's used at Opencaching.de. Just wanted to provide you with this information - please discard it if it's irrelevant.

harrieklomp:
Perhaps this is useless but i like to mention it. When playing and configuring GSAK i noticed that normally the  waypoint code OC**** is used for the link to a cache.
Because the OCPL sites works with "cache_id" and mostly not with "oc-waypoint"
For example click on mainpage on a cache (Newest caches) this shows a link with "oc-waypoint" when going to page http://www.opencaching.nl/newcaches.php and click on a cache it shows "cache_id".

kojoty:
@harrieklomp, I'm not sure if I understand you well. Do you need searchplugin script?

- If you want to see the cache by oc-waypoint use url: <site>/<oc-waypoint> - what redirect to <oc-site>/viewcache.php?wp=<oc-waypoint> 
For example http://opencaching.pl/OP352C -> http://opencaching.pl/viewcache.php?wp=OP352C

Generally I think that cache_id should not be presented anywhere outside the code - I think all url etc. should call cache by oc-waypoit... but it is for further refactoring...

Please also note that we want to save the url: /searchplugin.php?userinput=<oc-waypoint> as working after removing searchplugin.php script.

harrieklomp:
I do not need the searchplugin script :-)
I was only thinking out loud ;-)

andrixnet:
OC RO does not need it.

Thank you.

PR #485, opened 2016-04-08
mysql_* remove in context of #276 - next part

kojoty:


kojoty:
"Don't commit" because of weekend...

PR #489, opened 2016-04-09
consolidated InputFilter usage, prepared it for OKAPI

following5:
Submitted logs also need to be filtered in OKAPI; this change avoids redundant code when doing so.


following5:
Just noticed that not the InputFilter but the userInputFilter class should be used in OKAPI. log.php still uses InputFilter, therefore I missed that.

PR #518, opened 2016-04-13
mysql_* remove in context of #276 - next part - changes in util.sec

kojoty:
in context of #276


kojoty:
If anyone is able to test those changes It would be great because I don't have an idea how to be sure that all those changes works!

Issue #455, opened 2016-04-02
cache adoption by same user?


andrixnet:
I just discovered an old adoption request. Upon investigating, I gathered the following:
1. cache was published by another user
2. original owner offered the cache to me for adoption
3. my user adopted the cache (last year http://www.opencaching.ro/viewcache.php?cacheid=290)
4. my adoption page lists the same cache as both offered for adoption and waiting for adoption acceptance.

I tried to see what happens should I try to accept.. I got empty page and the following in the logs:
```
PHP Catchable fatal error:  Argument 1 passed to OrgCacheOwners::__construct() must be an instance of dataBase, instance of Utils\\Database\\OcDb given, called in /serve/www/oc_ro/oc_www/chowner.php on line 167 and defined in /serve/www/oc_ro/oc_www/lib/cache_owners.inc.php on line 9, referer: http://www.opencaching.ro/chowner.php
PHP Notice:  Object of class lib\\Objects\\User\\User could not be converted to int in /serve/www/oc_ro/oc_www/viewcache.php on line 1000, referer: http://www.opencaching.ro/chowner.php
```

I cancelled the adoption offer and point 4) above dissapeared. 

-----------------
Then I tried to send an adoption request explicitely to my username. 
The for accepted my input and I got the following situation: 
- cache offered for adoption in my list of caches
- nothing shows up in pending adoption (top of page).

I cancelled the adoption offer.

kojoty:
@andrixnet,
First: error you mentioned was fixed and shouldn't be observed anymore - if you got it - give me know please.

Second: I understand that the issue is in that the user can offer himself adoption by enter his own username as new owner name - right?

PR #526, opened 2016-04-14
Fixes #515, #467, #455

### LINKED ISSUE #515: Strange effects around watchers counter
Cache: http://opencaching.pl/viewcache.php?cacheid=49542

image

I click on "watch":
image

I click "unwatch":
image

I checked another cache and... it works! With this cache something is wrong...

**..and BTW: if every viewcache refresh should increase the counter of visits?**

kojoty:


kojoty:
@harrieklomp, @andrixnet 
This change needs one-line translation - please take a look at adopt_33

PR #530, opened 2016-04-15
fix fetch issue because of internal stmt

kojoty:
This is next issue because of one-instanced DB object (singleton) with internal statement...


kojoty:
I commit despite of weekend because this fix critical error

PR #573, opened 2016-04-25
new oc config idea

kojoty:
This is a "next" idea of new config for OC: 
- elastic configuration per module with: only defaults or defaults+local or defaults+node-specific or defaults + node-specific+local

- local file is not in repo and can not exists

- configuration files (defaults,node-spec., local) are classes (non-static)

- every module has it "Config" classes which:
  - extends CommonConfig class
  - checks and load proper class from set (defaults / node-spec / local )
  - provide static functions to access config settings

Currently there is a complete solution for two modules: Example and XExample. 
Base module is mandatory and must be present, because contains meta-information about configuration ( node-id which should be loaded and further "switch" between new- and old-way (based on settings.inc.php) of configs.

comments are welcome :)


andrixnet:
I'm not sure, but it seems, at least in part, tailored towards existing nodes quite explicitely, rather then in a generic way. 

Please also take into consideration this: https://github.com/andrixnet/opencaching-pl/blob/settings/lib/settingsDefault.inc.php

A configuration code can be used to parse (in a for loop) all keys at a certain level and go down levels (also modules) directly, then make those values available to the rest of the code. 
It could even build configuration items dynamically, based on the existing keys, without the need for specific code to be written for each variable.

kojoty:
@andrixnet, It's not that - solution I propose is an absolute generic solution, but with much more flexibility. Please note:

- we split configuration to separate independent modules
- files which contains configuration are located in /Config dir:
   - every module must has <moduleName>Defaults.php file with class  <moduleName>Defaults inside, and this file is always in git repo and is never locally (per node) overwritten
   - every module optionally can have a defaults specific for node in file <moduleName><nodeId>Defaults.php (with class <moduleName><nodeId>Defaults inside) which is stored in the git repo - it's allow to store non-critical part of config in repo for each OC node
   -  every module optionally can have a local overrides not stored in repo (here we store DB pass or google maps key etc.) in file <moduleName>Local.php with class <moduleName>Local inside

All classes inherit and overrides its content - and this is a simple and flexibility. Inside every class you can define properties and methods (methods seems to be a better option). 

Next every module needs Util to access its config from the rest of the code on easy way - and there is a place for <moduleName>Config files/classes. I moved it to the /Utils/Config for better split the "configurations" and "access to configuration".

Every <moduleName>Config provides static methods which return requested config var (see test.php code with examples). Internally <moduleName>Config stored singleton of proper class from /Config:
 - if there is only "defaults" class - it takes it
 - if there is "defaults" and "local" - it takes "local"
 - if there is "defaults", "node-specific" - it takes "node-specific"
 - ...and so on in manner locale before node-specific; node-specific before defaults 

and in fact this is the same as your idea from settingsDefault.inc.php but extended to more generic solution... I agree that it's not so simple as just 
`$config['server']['db']['port'] = '3306';`
but I believe that it can be much easier to maintain in long term...

If I can explain anything more give me know - my english is like it is, so maybe I couldn't describe it brightly :)

andrixnet:
What is the difference between "local" and "node specific" ?

andrixnet:
First : from syadmin point of view, it looks waaay too complicated. 
And I do mean to stress this point of view a lot. (my job is mainly sysadmin, though I do some web development on the side). 

Having a configuration file with a simple syntax is great. 
Having many configuration files is a lot more complex and headaches. 
Having configuration as programming code (declaring variables does not count) is a nightmare. 

Configuration files are meant to be simple. 
To be easy to maintain (without developer knowledge). Sysadmin is not meant to be (by definition) a PHP guru.
To be at least some level of syntax mistake tolerance. 

`$config['server']['db']['port'] = '3306';`
for me was inspired by MediaWiki like `$wgGroupPermissions['user']['edit'] = true;`

Such arrays can be dynamically parsed, even for deploying configuration code that needs not be written beforehand for each variable. 
On the other hand, if each variable is defined in the code and read from settings, then sanity checks should be performed on it. 

Many (most) MediaWiki config variables are defined with default values and only few are listed explicitely in the  config file, those which are absolutely required. 
All the rest can be customized if need be. They are well documented.

Think of it like this: 
I have to edit this file for module X, then that file for module Y then, oh, wait, to which module does setting Z belong, etc. Then: this is a local specific, this is a default, this is a node specific, for each of them. And make syntax mistakes, then oh, what did I do, where, wait, let me see the logs, etc, etc, etc. Oh, and I have to write PHP code for each setting. 
Honestly, my head is spinning already. 

**A well documented set of variables, some mandatory, some (most) optional. 
All have defaults defined according to repository. All can be overridden by sysadmin in settings file.
**
OC code is not that large and complex to go into such depth. 

On the other hand, there is at least a component that deserves modularizing in the truest sense (even independent of the repository): medals. 

I strongly urge against over complicating the configuration subsystem.
K.I.S.S.

kojoty:
@andrixnet, I understand your concerns but I think that your evaluation of complexity of this solution is too high. I agree that is can seemed to be too much complicated but I believe that it looks like that only at the first sight... but OK - the form of the config files is not so necessary to start working with it. 

Let's stay with the configs (settings.*) as is at this moment and add only "the intermediate layer" between the code and config files: simple access-to-config classes - nothing more at the first step.

It helps to make the settings format independent and allow us do anything what we decide in the future with configs :)  It's probably the same in function as what you called "glue". I prepare the new pull request soon.

andrixnet:
Please integrate (at least some of) my work here https://github.com/andrixnet/opencaching-pl/blob/settings/lib/settingsDefault.inc.php

Thank you.

kojoty:
sure!

PR #580, opened 2016-04-28
Removed HTML comments from inside <script> tags.

andrixnet:
Reason:
http://stackoverflow.com/questions/808816/are-html-comments-inside-script-tags-a-best-practice/808850#808850
http://www.javascripttoolbox.com/bestpractices/#comments

Removed language attribute from <script> tags.
Reason:
http://www.javascripttoolbox.com/bestpractices/#script


kojoty:
good work! :+1:

Issue #490, opened 2016-04-09
One interface for emails - code refactoring around emails


mzylowski:
Related with:
- #205 
- #238
- #48
Also:
- put all emails template in one directory
- create one object for emails and put all code for formatting templates into it
- transform all template into html style and put OC logo into it

harrieklomp:
For the last point it would be nice when the logo can be done via settings file.  This way every node can use their own logo.

andrixnet:
Or use logo defined for website top left (which is already personalized in settings file)

harrieklomp:
Perhaps a small suggestion. Remove one sitename. When someone register he/she sees twice the sitename in one line. For example this line "Welkom op Opencaching.pl Opencaching.pl!"

![selectie_009](https://cloud.githubusercontent.com/assets/1060518/14917968/ef8877b2-0e21-11e6-83a7-b61970987f7b.png)

mzylowski:
@up: done ;]

PR #587, opened 2016-04-29
Refactoring emails: part 1

mzylowski:
In the context of #490.

Waiting for review and suggestions.


lib/languages/en.php
```diff
@@ -1394,8 +1394,7 @@
     'runwatch13' => 'by clicking at',
     'runwatch14' => 'This message is auto-generated - please do not reply',
     'runwatch15' => 'No log entries',
-    'registermail01' => 'Hello',
-    'registermail02' => 'Welcome at',
+    'registermail01' => 'Welcome at',
```
andrixnet: Welcome to

andrixnet:
We should upgrade the HTML email templates as well to take into consideration display on smartphone clients. Currently OC generated emails look very small and almost always require the user to do additional zoom-ins.

lib/languages/en.php
mzylowski: Good catch, thanks!

Utils/Email/EmailSender.php
```diff
@@ -44,4 +45,32 @@ public static function adminOnErrorMessage(
...
+        $formattedMessage->setVariable("useruid", $uuid);
+        $formattedMessage->setVariable("server", ApplicationContainer::Instance()->getOcConfig()->getAbsolute_server_URI());
+        $formattedMessage->setVariable("sitename", ApplicationContainer::Instance()->getOcConfig()->getSiteName());
+
+        $formattedMessage->addFooterAndHeader($username);
+
+        $email = new Email();
+        $email->addToAddr($newUserEmailAddress);
+        $email->setReplyToAddr(ApplicationContainer::Instance()->getOcConfig()->getNoreplyEmailAddress());
+        $email->setFromAddr(ApplicationContainer::Instance()->getOcConfig()->getNoreplyEmailAddress());
```
kojoty: **in my opinion** this construction if far too complicated:
why to use create instance of ApplicationContainter to get instance of OcConfig?
Just use OcConfig::getNoreplyEmailAddress() method to get this field from config (if there is no such method just implement it)...

Something like ApplicationContainer can be a good solution as a base for Controllers and then provide config object, db object etc... but this is for the further discussion...

PR #508, opened 2016-04-11
Updated .htaccess for site aliases redirection moved to server config.

andrixnet:
Changes to .htaccess to make it independent of hostname.
Fixes https://github.com/opencaching/opencaching-pl/issues/105 for all current nodes.
OC NL = ready
OC RO = ready
OC PL = pending, new server and configs by @wrygiel

Issue #105, opened 2015-07-22
.htaccess redirection must be moved to Apache configuration


GoogleCodeExporter:
```
.htaccess file at root of site contains the following: 

RewriteCond %{HTTP_HOST} ^www\.(.+)$ [NC]
RewriteRule ^(.*)$ http://%1/$1 [NE,R=301,L]

This is a redirection rule to make http://www.website.XX redirect to this URL: 
http://website.XX

IMO this is a rather dirty way of doing it. It was also the cause of many 
problems with OCRO node that entered a redirection loop which I had a hard time 
figuring out. I commented out these lines, only to get SVN update conflicts 
when .htaccess was recently commited.

Since this affects the behaviour of individual nodes based on their preferred 
URL semantic, I suggest that such redirection be handled gracefully at website 
config level, rather then rewrite rules in htaccess. 

An elegant way to do it is the following: 

<VirtualHost *:80>
    ServerName www.opencaching.ro
[...]
</VirtualHost>

<VirtualHost *:80>
    ServerName opencaching.ro
    ServerAlias oc.opencaching.ro
    Redirect permanent / http://www.opencaching.ro/
</VirtualHost>

```

Original issue reported on code.google.com by `andrixnet` on 10 Apr 2014 at 12:25

andrixnet:
To all node admins: 
Please update Apache configuration such that this redirection can be removed from .htaccess file.

harrieklomp:
OCNL has also moved the config to the server itself.

mzylowski:
#508 reverted. Reopen.

wrygiel:
Summary of what was done (and it broke things):

This was removed from `.htaccess`:

```
RewriteCond %{HTTP_HOST} ^www\.(.+)$ [NC]		
RewriteRule ^(.*)$ http://%1/$1 [NE,R=301,L]
```

This was added to Apache:

```
<VirtualHost *:80>
    ServerName www.opencaching.pl
    Redirect / http://opencaching.pl/
</VirtualHost>
```

It seems that the first piece (the one which currently resides in `.htaccess`) doesn't work in `/okapi/` directory. Any idea why?

deg-pl:
We must have "normal" VirtualHost entry with DocumentRoot, Directory, Options etc.
Remove "Redirect / http://opencaching.pl/"
Add:
```
RewriteCond %{REQUEST_URI} !^/okapi/
RewriteRule (.*) http://opencaching.pl$1 [R=302,L]
```

[...]

PR #589, opened 2016-04-30
Improved ocpl-update.extended.sh (update script).

andrixnet:
* Features:
- OPENCACHING-PL website files owned by designated (non-root) unix user
- select branch to sync to (default: master)

* Sysadmin options:
- log each update to syslog
- select log facility (default cron.info -> /var/log/cron)
- keep local changes between updates, if really necessary
- show modified files by update operation

### DIFF
--- a/docs/system/ocpl-update.extended.sh
+++ b/docs/system/ocpl-update.extended.sh
@@ -4,48 +4,81 @@
 # the following script:
 # https://github.com/opencaching/opencaching-pl/blob/master/post-commit.php
 
-# site name, FQDN, not URL
+###########################################################################
+# Site configuration
+###########################################################################
+
+# site name (FQDN)
 SITE_NAME=YOUR_OPENCACHING_HOSTNAME
+
 # hosting root directory
 SITE_ROOT=/path/to/OPENCACHING-PL
+
 # user owner of the files.
 # must have primary group same as webserver user
-SITE_USER=oc_user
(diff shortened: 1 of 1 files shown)


mzylowski:
I'm no so much experienced in ocpl automation mechanism to review this PR.
Somebody else? ;]

kojoty:
@andrixnet can you explain in a few words how/when/why we uses this script?

andrixnet:
This is an administrative extension to the basic update script suggested by wrygiel after the move to github.

It allows: 
- simple update of the website
or
- configurabile administrative actions: 

1. log to syslog on every update; configurable log channel and level
Allow sysadmin to see update history and potentially track problems with the updates on that server
2. option to protect local changes from being overwritten. 
Example: preserve modified .htaccess until #105 and #508 are solved
3. option to include list of modified files when logging each update.
4. perform all actions in a unified configurable way. Only need to edit config variables.

I have applied a similar script with SVN while at google code and it helped me greatly (as sysadmin) to track down and fix some inconsistencies and locally broken update process.

It also makes good system administration practice, to keep track of changes to your server.

andrixnet:
One needs to set SITE_NAME, SITE_USER, REPO and BRANCH to proper values.
Having DO_LOG=0 and KEEP_LOCAL=0 makes the script behave like the simple update script as first deployed with github.

PR #614, opened 2016-05-11
Zmiany w forum/wiki na https

deg-pl:


andrixnet:
shouldn't these references be from the array of wiki links defined in config?

deg-pl:
Yes. You're right. This was a quick fix ;)

PR #667, opened 2016-06-03
googlemapapi ver update

kojoty:
now we can use ver. 3.24 (current released version)


kojoty:
commit waits for Monday

Issue #611, opened 2016-05-10
Remove events from 'cache found' count. Recalculate user stats.


mzylowski:
For me (and a lot of others opencaching users) including events into cache founds it's bad idea. 
Event isn't a cache.
We talk about it in #593, we receive e-mails like that:
`Cześć.
Czy te spotkania zawsze były doliczane? Spotkanie to nie skrzynka jej się nie szuka. Więc w statach spotkań niech pokazuje na ilu spotkaniach byłem  a w statach ogólnych tylko znalezione Kesze.
Pozdrawiam`

In this issue:
- we need changes in code
- run script for recalculate stats

andrixnet:
Can events be counted separately and shown? 
or
Can "attended" count as a "found" on per node basis?
OCRO prefers it like this and I know OCNA would like to rejoin within the next year the current OPENCACHING-PL code, and this was one specific issue of theirs.

harrieklomp:
For oc.nl it is logic that events are no caches, so should not be counted as found cache.
There are some parts in the code who does count them as found and some show them separate.

For example on http://www.opencaching.pl/viewprofile.php?userid=?? show only caches as found and events are shown separate.

deg-pl:
We have a lot of users surprised counting events to "Number of found caches". Another mail today:

> mam pytanie odnośnie statystyk na OC.
> Wiem, że sporo działacie ostatnio ale wychwyciłem małą nieścisłość pomiędzy ilością znalezionych skrzynek z poziomu mojego profilu (4384 na chwile obecną) a ilością znalezionych skrzynek z poziomu ogólnych statystyk (4368 na chwile obecną). Różnica to 16. 
> Skąd to się bierze, ktoś już może coś takiego zgłaszał, itd? W wolnej chwili mógłbyś zerknąć czy pisać oficjalne pismo do RT? ;-)

Opencachingna:
We use the old OCPL code at opencaching.us. This code did NOT count events as finds anywhere on the site. However, we had several requests over the years that they DO count as finds. So I went in to I'll bet about 15 files on the site, and changed the code so they do count. This must be a North American thing that people want them to count as finds. :-)

harrieklomp:
@Opencachingna I think this is more a GC thing. Some things should not be the same as on GC, for example this count a event as found :-)

Opencachingna:
[...]

PR #671, opened 2016-06-06
Tymczasowe rozwiazanie problemu statystyk zliczajacych uczestnictwa w…

deg-pl:
… eventach do znalezien


mzylowski:
@deg-pl You shouldn't just change translations without info to @harrieklomp or @andrixnet 
You also should change English translation.
I know it its temporary commit but we have to keep all translations unified every time.
I understand - we are not native English speakers but commit message in polish? It is a deterrent from our repo :P Harrie even doesn't know which issue is related with this commit and why this change was made.
#611

PR #741, opened 2016-08-31
Partial Czech translation

anomen-s:
This is partial translation to Czech. Remaining texts are in English


mzylowski:
Great job!

Issue #769, opened 2016-09-20
Google Analytics key should be read form config and GA should be load for all pages


kojoty:
We used GA in many scripts right now but pages from all running versions of the GA code are counted a oc.pl visits in GA.

GA key should be set in config to allow OC node administrator enable it for its nodes.

kojoty:
GA for main site is fixed - we still need changes for "mobile" version of OC ...

andrixnet:
After reading on the mailing list, I was on my way to investigate and report it as an issue.
Good job, @kojoty !

This solves https://github.com/opencaching/opencaching-pl/issues/300

GA for mobile version is opened here: https://github.com/opencaching/opencaching-pl/issues/239

deg-pl:
Can we close this issue?

PR #770, opened 2016-09-20
#769: GA key is loaded from config and GA is used for all templates

kojoty:
change in context of #769


kojoty:
@harrieklomp, @andrixnet This change allow you to decide if you want to track your site visits with GA or not (till now visits at your oc nodes GA counted as visit at ocpl website because our GA key was hardcoded in code)

If you would like to enable GA for your sites you need to obtain GA key from google and set it in settings.inc.php in var: $googleAnalytics_key (settings-example contains the example).

Another improvement is that now all pages (excluded xml interface) should be tracked by GA - only pages use main.tpl was tracked before.

PR #778, opened 2016-09-26
file uploads look&feel

DorotkaMakotka:
example: 
fileupload1


DorotkaMakotka:
@harrieklomp , @andrixnet:
Hi, please note that there is a small translation change :) thanks!

andrixnet:
Will update translation in a larger batch. Thank you very much for the notification. 

IMO the "browse" button would be better if it would be visually different then the action buttons. 
I say this because the action button for a form with a file upload depends on the browse button being pressed first. So a visual cue to direct the user to that button first would be nice.

Thank you.

andrixnet:
Translations updated in https://github.com/andrixnet/opencaching-pl/commit/b368bec0551354547b32c7708cdc62870825dddd

Issue #790, opened 2016-10-26
Adaptation OC PL code to Opencaching UK


deg-pl:
This is a technical issue - joining all commits related to new OC UK node.

andrixnet:
Please take this https://wiki.opencaching.eu/index.php/Cache_types into consideration if there are any adaptations needed for cache_types. AFAIK OCUK had "educache" type. If so, please make the adaptation such that "educache" gets type 17 as suggested.

deg-pl:
@andrixnet OC UK team decided to convert educache to virtual cache.

deg-pl:
@wrygiel please add https://opencache.uk/post-commit.php to post commit server list on Github

andrixnet:
@deg-pl there is a problem (I don't remember if I got to report it) : someone published virtual cache (as alternative to earthcache on gc.com) and it got transformed into unkown after submission. 
http://www.opencaching.ro/viewcache.php?wp=OR0172
http://www.opencaching.ro/viewcache.php?wp=OR0170

I checked just now: I can't hide a virtual. While filling the form I forgot some fields, and even the error taking me back to the form already transforms virtual into unknown. 

You have my settings file, according to it, virtual cache type is allowed.

andrixnet:
Note: translation of geopath (and others is a work in progress. should be done by the end of the weekend)

wrygiel:
> @wrygiel please add https://opencache.uk/post-commit.php to post commit server list on Github

Done!

andrixnet:
see https://github.com/opencaching/opencaching-pl/issues/663

andrixnet:
opencache.uk back on wiki.opencaching.eu main page.

andrixnet:
Links page needs updating : https://opencache.uk/articles.php?page=links

On homepage, static map is empty. 

On cache page, static map is empty.
Popup map loads for a moment, then shows the error like on the geopath pages.

On geopath pages : map does not load.
> This page didn't load Google Maps correctly. 

Same for Map v3

andrixnet:
@deg-pl could you please add the UK attribute images to the opencaching-pl tree so that other nodes may also use them?

andrixnet:
wiki.opencache.uk ERROR while uploading file : 

> Exception encountered, of type "LocalFileLockError"

andrixnet:
Translation help, please (both for wiki and strings): what would you say is the best term for findind all caches in a GeoPath? 

1. gained (current version, sounds bad)
[...]

PR #800, opened 2016-10-31
Added Google keys to JS API, removed obsolete sensor parameter and fi…

deg-pl:
…xed errors with fixed mode in charts (#790)


andrixnet:
OCRO updated with new key.

Issue #120, opened 2015-07-22
region from coordinates yields nothing


GoogleCodeExporter:
```
When creating or editing a cache, the button "region from coordinates" opens a 
new window to show the region where these coordinates are located, listing 
country, region as found by OCRO and region as found by Google Maps. 

In the case of OCRO, I always get an empty list. 

See attached image.

```

Original issue reported on code.google.com by `andrixnet` on 3 Jun 2014 at 1:12

Attachments:
- [ocro-region.png](https://storage.googleapis.com/google-code-attachments/opencaching-pl/issue-120/comment-0/ocro-region.png)

kojoty:
@andrixnet , is this still an issue? Maybe we can close it?

andrixnet:
Will have to reinvestigate. 
However my last assessment was that database contents should be redefined such that insteand of `LINESTRING`, shapes are stored as `POLYGON`.

kojoty:
It seems that (thanks to Deg and OCUK action) I made progress in this subject. I have a new content for nuts_* tables with update to NUTS-2013 and now nuts_layers 
[nuts_codes.sql.zip](https://github.com/opencaching/opencaching-pl/files/563995/nuts_codes.sql.zip)
[nuts_layer.sql.zip](https://github.com/opencaching/opencaching-pl/files/563996/nuts_layer.sql.zip)

contains POLYGONS instead of LINESTRING. We still need tests and fixes (for example mysql WITHIN is now MBRWITHIN - we should use ST_WITHIN).

deg-pl:
Thanks for your work!
We also should do:
`UPDATE `nuts_layer` SET `code` = REPLACE(`code`, 'UK', 'GB') WHERE `code` LIKE 'UK%';
UPDATE `nuts_codes` SET `code` = REPLACE(`code`, 'UK', 'GB') WHERE `code` LIKE 'UK%';`

kojoty:
I'm not sure if UK == GB: [(http://brilliantmaps.com/eng-gb-uk/)] - are you sure it works OK?

deg-pl:
Problem has source into official ISO code for United Kingdom. ISO code is... GB, not UK. This is source of differences...

kojoty:
I prepared the new-better versionn of nuts_codes table and updated link above - here are good files for NUTS-2013 again:
[nuts_codes.sql.zip](https://github.com/opencaching/opencaching-pl/files/567865/nuts_codes.sql.zip)
[...]

PR #819, opened 2016-11-04
Correction of PL region names - related to #120 and #147

deg-pl:
@andrixnet @harrieklomp import this SQLAlter AFTER import of new nuts_* tables.
This is not very important for not OC PL sites.

Done at OC PL and OC UK


harrieklomp:
On OCNL these Polish regions are imported.

andrixnet:
Applied on OCRO

PR #845, opened 2016-11-22
On viecache: disable OSMapa, enable OSM. Small language correction

deg-pl:
OSMapa currently doesn't work with zoom lower than 13. I disabled default use OSMapa as link in viewcache. I enabled OSM instead of OSMapa.

This could be easu overwritted in setting.inc.php by:
```

$config['maps']['external']['OSM'] = 0;
$config['maps']['external']['OSMapa'] = 1;

```
(but I think that it has no sense).

@harrieklomp - your settings.inc.php doesn't include settingsDefault.inc.php. If you want to enable OpenStreetMaps on viewcache pages, you should include in your settings.inc.php:
```

$config['maps']['external']['OSM'] = 1;
$config['maps']['external']['OSM_URL'] = '<a target="_blank" href="http://www.openstreetmap.org/index.html?mlat=%1$f&mlon=%2$f&zoom=16&layers=M">%6$s</a>';

```


deg-pl:
@andrixnet - you have $config['maps']['external']['OSMapa'] = 1; in your settings.inc.php. Please consider to disable OSMapa by setting this to 0.

harrieklomp:
Enabled OSM on OCNL.
Thanks @deg-pl for your attention and notification.

Issue #851, opened 2017-01-10
There still is possibility to set WIS < 66%


mzylowski:
I don't know how but somebody done this today.
On oc.pl minimal WIS for geopath is 66%

![bug](https://cloud.githubusercontent.com/assets/13780868/21812507/752d674c-d753-11e6-91ce-3cc305d7848e.png)

![bug](https://cloud.githubusercontent.com/assets/13780868/21812572/b2109c6a-d753-11e6-8348-9a845b3db162.png)

Probably we should check value when we saving to db, and when is lower then minimal WIS we should force it.

After fixing this issue we should fix all geoPath in db.

Reproduction: http://opencaching.pl/powerTrail.php?ptAction=showSerie&ptrail=548

cc @wloczynutka

wloczynutka:
There is minimal demand percent validation on backend:

https://github.com/opencaching/opencaching-pl/blob/master/powerTrail/ajaxUpdateDemandPercent.php 

line nr. 14.

uses PowerTrailController::MINIMUM_PERCENT_REQUIRED which is hardcoded in

https://github.com/opencaching/opencaching-pl/blob/master/lib/Controllers/PowerTrailController.php
line nr. 13

So somebody had to used some other script (?) maybe via COG?

wloczynutka:
No, via COG impossible.

If server is colecting POST logs for script ajaxUpdateDemandPercent.php 
maybe you can obtain some from there, (see what was really posted). 
Otherwis it may be difficult to realize what kind of exploit was there. 
(direct db change?)

wloczynutka:
One place I did not checked: creating new geo path function

wloczynutka:
checked: creating new geo path function, there is bug. This function allow any value beetwen 0 and 100.

mzylowski:
I will close it after all db updates.

PR #852, opened 2017-01-11
Minimal geoPath percent value

mzylowski:
Value is changed from 66.6 to 67 because we parsing value to int in the backend. Also in wiki rules is 67% 
https://wiki.opencaching.pl/index.php/Geo%C5%9Bcie%C5%BCka#Wymagana_ilo.C5.9B.C4.87_skrzynek_czyli_.22WIS.22

After merging it I will change WIS manually in database (for geopathes with wrong values)  (after communication with COG and owners).

### DIFF
--- a/powerTrail/powerTrailController.php
+++ b/powerTrail/powerTrailController.php
@@ -200,6 +200,9 @@ class powerTrailController {
         {
             $query = "INSERT INTO `PowerTrail`(`name`, `type`, `status`, `dateCreated`, `cacheCount`, `description`, `perccentRequired`) VALUES (:1,:2,:3,NOW(),0,:4,:5)";
             $db = OcDb::instance();
+            if ($_POST['dPercent'] < \lib\Controllers\PowerTrailController::MINIMUM_PERCENT_REQUIRED) {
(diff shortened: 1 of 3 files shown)


powerTrail/powerTrailController.php
```diff
@@ -200,6 +200,10 @@ private function createNewPowerTrail()
         {
             $query = "INSERT INTO `PowerTrail`(`name`, `type`, `status`, `dateCreated`, `cacheCount`, `description`, `perccentRequired`) VALUES (:1,:2,:3,NOW(),0,:4,:5)";
             $db = OcDb::instance();
+            if ($_POST['dPercent'] < \lib\Controllers\PowerTrailController::MINIMUM_PERCENT_REQUIRED)
+            {
```
andrzejwozniakmorele: [psr-2](http://www.php-fig.org/psr/psr-2/)  powiada, że klamerka powinna być w tej samej linii co if

mzylowski:
Was related with #851

andrzejwozniakmorele:
maybe instead changing WIS manualy it would be beter finish work on automatic geopath guard?

wloczynutka:
maybe instead changing WIS manualy it would be beter finish work on automatic geopath guard?

PR #871, opened 2017-02-22
#782: sql changes - first part

kojoty:
Hi @andrixnet, @harrieklomp, @deg-pl 

I prepared changes around #782 (and a full refactoring of the code from viewcache.php)
To deliver it all I need to update DB at all our nodes.

Please give me know here when your nodes will be ready.

### DIFF
--- a/sqlAlters/2017-02-23_cache_visits_refactoring.sql
+++ b/sqlAlters/2017-02-23_cache_visits_refactoring.sql
@@ -1,0 +1,22 @@
+-- 2017-02-20: Refactoring of cache visits
+-- @author: kojoty
+
+-- This is the first stage of changes: new table (cache_visits2)
+
+
+CREATE TABLE `cache_visits2` (
+  `cache_id` int(11) NOT NULL,
+  `user_id_ip` varchar(15) COMMENT 'user_id or used IP address',
+  `type` varchar(1) NOT NULL COMMENT 'C=cache_visits; U=last_user_unique_visit; P=prepublication_user_visit',
+  `count` int(11) NOT NULL DEFAULT '0',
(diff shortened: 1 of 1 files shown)


deg-pl:
Done @ OC PL and UK

harrieklomp:
Done on OC NL.

andrixnet:
Unable to apply at OC_RO due to error : 
`ERROR 1067 (42000) at line 7: Invalid default value for 'visit_date'`

My sql_mode variable is empty (NO_ZERO_IN_DATE,NO_ZERO_DATE are not set) : 
```
show variables like 'sql_mode' ;
+---------------+-------+
| Variable_name | Value |
+---------------+-------+
| sql_mode      |       |
+---------------+-------+
```
Running on 5.5.54-MariaDB according to distro stable version at latest security update.

The problem lies in the fact that column type is set to DATETIME which does not accept this default value. TIMESTAMP however accepts this default value with no problems. 
The difference is that DATETIME also stores decilams for seconds, while TIMESTAMP does not. 
MariaDB accepts this default for DATETIME starting with 10.0.1, which is not the case here. 

Is it ok to use TIMESTAMP column data type?

kojoty:
@andrixnet, It seems we used datetime in many other places in db so maybe let's stay with this type - I'm not sure if it can make any difference in future. 

Please skip default value for this field and I'll remove default value from the definition of this table in sql-alter in next stage of this change for the rest of our oc-nodes.

andrixnet:
Very well, please note however, if you use DATETIME and no such default, the code must explicitely set the value for this field.

SQL changes set on OCRO, without the default value as noted above.

kojoty:
@andrixnet 

> ...please note however, if you use DATETIME and no such default, the code must explicitely set the value for this field.

sure - I know about it :)

PR #885, opened 2017-02-27
#782 visit counter

kojoty:
Hi,

This is the main part of my current works around viewcache. The excuse for this change was cache-visit-counter from #782 but changes are much more wide and this is complete refactoring of viewcache controller + mass cleanup of viewcache.tpl.php + many small fixes and improvements around it.

Please note: I've tried to test this change as much as I could but there could be and probably are some issues - please give me know and I tried to fix it ASAP.

@harrieklomp , @andrixnet - there are small translation changes - please take a look at this.

The main changes, which I would like to consult with you:
- **Controllers dir**: I believe it's time to start step-by-step refactoring to the model with controller which rule the user request + model (object) which exchange the data with DB + View which contains HTML and decide how the data is presented. I think it is a good idea to store controller classes in Controllers dir. There are still stub in /viewcache.php because we don't have a good router class - but this is temporary...
Please take a look and review.

-  **Viewcache template changes** are actually small and not final - I'm not an UI designer - I can't make it as beautiful as I would like but at least now it is much more easy to change this view - all HTML code is inside one file, there is a separate css for this file only and much more... - please take a look and if you have any idea how to improve the look of viewcache just do the change or give me know...

- **cache visit stats** - be warned that all cache visit stats after this change are zeroed - old stats lost any significant information because every refresh of the visitcache increments its value


harrieklomp:
:+1: 
Translation will be done after your commit for OCNL

andrixnet:
same for OCRO.

kojoty:
@deg-pl, @andrixnet, @harrieklomp 
Pull request merged. Please note that there is an small sql alter which remove default value in cache_visit2 db table.

kojoty:
@andrixnet, @harrieklomp 
one more thing: there was a change of external maps links in settingsDefault.inc.php ( $config['maps']['external'][ ... ]) - if you have any change of defaults for external maps in your OC-node config (settings.inc.php) - it needs to be updated.

deg-pl:
sqlAlters/2017-02-27_cache_visits_refactoring_part2.sql done at OC PL and OC UK

PR #893, opened 2017-03-01
andrixnet patch for nano-size support + minor cleanup around cache sizes

kojoty:
@andrixnet 
Here is your work from [here](https://github.com/andrixnet/opencaching-pl/commit/22f2263586211fdbdeb746fe2fa2900633205afa) additionally my minor cleanup around size translations. Please review if everything is like you provide before and we can merge this pull request.

@harrieklomp , @deg-pl 
there is a small DB change, which should be applied to your nodes


harrieklomp:
Changes in OCNL DB are done.

deg-pl:
We have incompatibile structure of the table cache_sizes (at OC PL and OC UK):
CREATE TABLE `cache_size` (
  `id` int(11) NOT NULL,
  `pl` varchar(60) NOT NULL,
  `en` varchar(60) NOT NULL,
  `nl` varchar(60) NOT NULL
) ENGINE=MyISAM DEFAULT CHARSET=utf8;

andrixnet:
At OCRO: 
```
CREATE TABLE `cache_size` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `pl` varchar(60) NOT NULL,
  `en` varchar(60) NOT NULL,
  `nl` varchar(60) NOT NULL,
  `ro` varchar(60) NOT NULL,
  PRIMARY KEY (`id`)
) ENGINE=MyISAM AUTO_INCREMENT=9 DEFAULT CHARSET=utf8
```

andrixnet:
Note: OCRO is currently on branch https://github.com/andrixnet/opencaching-pl/tree/well_defined 

I have done some changes to allow SIZE_NANO and SIZE_OTHER to work properly and found out at least 3 places where there had to be defined. (since then clarified for future development with LZA)
There is a snag, however: 
$config['forbiddenCacheSizes'] only applies properly on editcache. 
Doesn't work on newcache and search pages. 
Not sure about the others. 

Code in lib/cache.php has to be rewritten into lib/Objects/GeoCache.php

If these types are forbidden in config, several more changes might be required to actually and properly apply these restrictions. 
I've only realised this today.

andrixnet:
There is another mitigation: 
1. when a size (ie nano) is forbidden, it cannot be selected for new caches
2. what happens to caches of that size before it was forbidden? 
- edit cache forces a size change?
- edit cache keeps size, but does not allow other caches to be set to that size
- does search page allow search by that size? 
(assuming caches of sizes other then the currently allowed have been allowed in the past)

I think these items need clarification.

andrixnet:
viewcache should not be affected, but search page and geopath might be, in the grounds that forbidden cache sizes are allowed or not to be searched and displayed as separate criteria.

harrieklomp:
1. On OCNL user cannot select micro caches when making a new cache.
2. On OCNL we have some micro caches. They where made before it was forbidden. At this moment we have the following.
We cannot force a user to change the size of his cache but when edit his own cache it is possible to change to bigger size. When a micro cache is changed to ie. small and saved, it cannot changed back to micro.
As long there are micro caches on OCNL the user have the right to search for them on the search page.

BTW, On the search page the size "Micro" is read in the old Database (i have changed it in XmicroX for today on OCNL).

deg-pl:
OK. I've done first:

> ALTER TABLE `cache_size` ADD `ro` VARCHAR(60) NOT NULL AFTER `nl`;
> UPDATE `cache_size` SET `ro`=`en` WHERE 1

and next your sqlAlter.

harrieklomp:
Reverted changes on OCNL Database. Because cache size "Nano" and "Not specified" are shown on searchpage.

deg-pl:
The same @ PL and UK

harrieklomp:
Another BTW: Edit a cache is also read from original Database. My change in DB to XmicroX is visible when edit a cache. This is only possible for admins, for normal users this micro is not visible.
selectie_104

andrixnet:
Due to all these side effects, I think it would be best if these changes were not integrated at this time. 
They work when they are not forbidden, they don't work properly if they are forbidden. 
When the rework of the data types is complete, these must not happen anymore.

kojoty:
OK, let me understand what in fact we need:
- config should provide list of allowed sizes for new caches
- config should provide list of existed sizes of caches (which can be different than allowed)

and then on new-cache + edit-cache pages - size can be set only from "allowed sizes"
and on search page all values of "existed sizes" should be presented

Is that all we need?

@andrixnet - we need this "nano" because of integration with OCNA, right?

harrieklomp:
Why not add all sizes in code.
In settings file we have 'forbiddenCacheSizes' => array(..... and with this each node can disable cachesizes.
The searchpage can use this also.

andrixnet:
That is my overall goal here https://github.com/opencaching/opencaching-pl/projects/2

kojoty:
I close this PR - if we come back to this case we can open it again...

PR #925, opened 2017-03-23
Added git revision to OC footer

mzylowski:


andrixnet:
Is this really necessary? 
IMO it requires too much CPU usage by external command execution to be justified, and do it for each pageload. 

If it is deemed really necessary, then I would recommend doing it via the update script, set a value in sysconf table, then read it at runtime. IMO much faster.

kojoty:
@mzylowski, I also afraid if it is a step in right direction: 
- I'm not sure if we need this - most of oc users doesn't need to know the current version of the service code - anyone asked for this?
- updates of the production sites are automatic so we be sure that if something is pushed to master it is present on the server (fail is really rare)
- this code needs to be executed for every display of the site what costs (not much, but this is a cost)

BTW. I think we can provide some kind of "changelog" as a separate page at OC site, which can contain list of commits names (we can decide to put some 'tag' in commits name to distinguish which commits are important) - I believe it could be interesting for users.

mzylowski:
> fail is really rare
But sometimes.. ;p
> updates of the production sites are automatic
Yea, but this should be helpful for example on our develop VM's.
> IMO it requires too much CPU
This convinces me.

Issue #917, opened 2017-03-16
GPX: XML Parsing Error: not well-formed


andrixnet:
Please visit here: http://www.opencaching.ro/search.php?queryid=1465&startat=0 and export GPX file for all results. 

Try loading the GPX file into an XML parser. 
I tried in Firefox (plain parser), GSAK, GPSBABEL. 
Each give the following error: 
`XML Parsing Error: not well-formed`

Firefox shows this error: 
```
XML Parsing Error: not well-formed
Location: file:///E:/download/OC.gpx
Line Number 15, Column 47:                <desc>[LALEA] Baraj lacul Gole￿ti de la andrixnet, Traditional Cache (2/2)</desc>
----------------------------------------------^
```

A closer look at the generated file shows the header being properly generated : 
`<urlname>Opencaching.ro - Geocaching ├«n Rom├ónia</urlname> ` 
_(this is by an ASCII viewer)_

` ├«n ` = 20 **C3 AE** 6E 20
Properly shown as UTF-8 character  = **în**
This is the correct testcase.

=================================
`<groundspeak:name>[LALEA] Baraj lacul Gole┼ti</groundspeak:name>`
`e┼t` = 65 **C5** 74
Proper UTF-8 character would be = **eşt**
Correct byte sequence would be : 65 **C5 9F** 74 
Thus it is apparently an incomplete byte sequence. 

The problem goes on for the rest of the file.

kojoty:
Hi @andrixnet,

few ideas at first sight:
- there wasn't any significant change in source code around gpx production so it not a bug in a code (probably)
- it seems that something is wrong with encoding of texts from DB - non-ascii chars are in different encoding
- it also seems that this is specific to ro node - I can't reproduce similar issue with my local dev-node
- I assume it works before - right?
so:
1. have you check if you have any warnings in apache error log - maybe we find some clue there - can you send me your error.log from today?
2. what is a encoding in your DB? - can you send me a dump with structure of table cache and cache-desc?
3. can you identify when it worked last time - maybe you upgraded software on server or something?

andrixnet:
We've done a lot of content translation and proper localization in the recent past, so, let's say this time last year, we had a lot less such characters. 

Nothing in the error log. 
Only one thing appears for KML:  `PHP Notice:  Undefined variable: kmlDetailHead in lib/search.kml.inc.php on line 165` but it's not even GPX related.

[...]

PR #928, opened 2017-03-25
Add translatio strings to GCT

andrixnet:
Fixes https://github.com/opencaching/opencaching-pl/issues/924

I know it may not be the most elegant solution, but apparently works :-)

PR #931, opened 2017-03-25
GPX language conversions and code cleanup

andrixnet:
While trying to fix https://github.com/opencaching/opencaching-pl/issues/917 I encountered several inconsistencies and i tried to resolve them as best I could. 

Most search.*.inc.php defined the same functions to do some work for them. I moved those functions to common.inc.php with several improvements. 

myroutes_search.php uses the same function names, but slightly different implementation. For now I renamed them to isolate them. Can this be merged somehow with the search.*.inc.php borne functions? 
A similar situation is in ocxml11.php. I know this is a sensitive API to older apps, so I took the same action as for myroutes_search and ask the same question. 

Is the OVL format used ? 
I found a syntax error in search.ovl.inc.php that was left there probably since DB refactoring over half an year ago. 

I kept Plconvert (renamed as conver_lang) but didn't use it. 
I found some code that makes use of iconv, which I consider a much better choice then a manual replacement list (which would work fine for a single language). 
**I replaced the use of convert_lang with convert_text (uses iconv)**
This must be tested in other languages, I only tested with romanian content. Seems fine. 

I made some changes to loadlanguages with setlocale (ro was missing, added upcoming hu and added LC_CTYPE, recommended for iconv usage).

---
Please review my changes and test them in your development machine with other language data then english and romanian.

Thank you. 

P.S. 
future note to self: these file exports redefine a whole lot of data definitions (like cache types, cache sizes, etc), so they will have to be heavily reworked for https://github.com/opencaching/opencaching-pl/projects/2


lib/js/GCT.lang.php
```diff
@@ -0,0 +1,22 @@
...
+ * /lib/js/GCT.js
+ * ************************************************************************
+ */
+ 
+$rootpath = '../../';
+require_once $rootpath . 'lib/common.inc.php';
+
+header("Content-Type: text/javascript");
+
+echo("
```
kojoty: I think much more readable and easier in maintenance is to close php in such situations for example here you can (probably) just do:

_(...)
?>

GCT_lang = {
  prev: <?=tr('pagination_left')?>,
  next: <?=tr('pagination_right')?>,
  number_of_caches: <?=tr('number_of_caches')?>,
 }_

kojoty:
Hi Andrei,

Generally this is a good idea, but I have a few opinion about this:
1. In my opinion we should avoid adding anything to the common.inc.php - actually I would like one day to remove this file at all but there is still too much important code - of code is needed by some subset of scripts it should be moved to separate library and included by this subset of scripts - not by all of the scripts - so in this case you can consider to move this functions to the file called for example search.common.inc.php and include this file wherever it is needed. Alternatively if I remember well all search*.inc files are loaded only by search.php so maybe you can put this functions to the search.php - then they should be accessible from all search.*.inc.php scripts

2. Do you know why we need to convert from utf-8 to ascii for gpx files? - I don't know much about gpx and maybe this is reasonable but I also would not be surprised if this conversion is not necessary at all. Do you have some knowledge about it?

another comments in the code...

mzylowski:
I will try test your change on my devel in tomorrow evening

andrixnet:
@kojoty 
I moved those to common.inc.php for the reason that they were used in lib.search.*.inc.php, myroutes_search.php and ocxml11.php (which still needs clarification). Not only as included from search.php
Of course, it can be another library file. another library can be though of. For example there are many GPX definitions that are redefined in each place they are used, as @harrieklomp pointed out to me. 

Converting from UTF-8 to ASCII already existed in the code (though not working properly). 
Reasons for which to do this are third party applications and devices that can import GPX files but cannot handle or display UTF-8 characters properly or they simply don't have fonts with sufficient characters. One such example is GSAK. 

Since everything is UTF-8 in the website, and I assume everything else in the site works properly, then the data is consistent. 
I would not consider GPSBABEL for such a conversion, because it will lose fields deemed unnecessary by it. 

I would consider however a variant to offer GPX with UTF-8 content and converted GPX because after export is quite hard for the average end-user to make such conversions. 
(though I would also suggest filtering the resulting GPX through iconv utility on the server rather then dump the conversion on the user)

lib/js/GCT.lang.php
mzylowski: @kojoty Your sugestion from here is solved by andrie via #928

mzylowski:
@andrixnet After merging #928 here will be a conflict. To avoid, please remove cdf4446 from this PR.

About the rest: LGTM, except kojoty comments.
@kojoty Merge button in your hands.

This should be merged after #928, so for now 'Don't commit label'.

@andrixnet thanks for this PR ;]

andrixnet:
Keep Don't commit label. I have further improvements on the way.

lib/languages/pl.php
```diff
@@ -1118,6 +1118,8 @@
     'route_03' => 'Wyczyść trasę',
     'route_04' => 'Reset mapy',
     'route_05' => 'Wczytuję trasę ...',
+    'route_06' => 'Maximalna ilość logów musi być cyfrą!',
```
andrixnet: Please confirm this translation. 
Also please confirm the meaning of the string. 

(old meaning said "minimum" in a context about a "maximum"

andrixnet:
@kojoty @mzylowski  @harrieklomp 
Please take a look at this. 
I've just finished working the file exports. 

Sorry for the one huge commit, I worked for many days for short periods between other work and I couldn't manage to assemble changes atomically. 
For a log of the changes please look mainly at https://github.com/opencaching/opencaching-pl/pull/931/commits/da93bd9e3201a64f36f6bd866362157f4f3466b2

Main improvements: 
- OC generated GPX files now are proper, albeit converted to ascii. Fields are consistent. Attributes are properly set. String language is consistent. ZIP is disabled (as it was set before)
- KML now has improved icons and consistent data
- opencaching.kml live Google Earth exporter is now generated using your profile's home coords or site home coords.
- Google Earth exporter uses same styles as KML file export. 
- KML styles based on c:geo marker_oc graphics
- WPT, TXT, UAM, OVL, OVM, LOC files improvements. 
- removed unused files
- removed unused bits of code.

A few translation changes and a few hardcoded strings translated. Please check.

andrixnet:
If there are no objections, may it be merged after this weekend?

kojoty:
@andrixnet, From my side - I'l try to review it tomorrow.

lib/languages/pl.php
kojoty: @andrixnet: better polish translation could be probably:
'route_06' => 'Maksymalna ilość logów musi być liczbą!',
'route_07' => 'Dozwolona wartość maksymalnej ilości logów musi być z zakresu: 0 - 999',
(you can change it or I'll fix it later)

lib/js/GCT.lang.php
```diff
@@ -0,0 +1,22 @@
...
+var GCT_lang = new Array();
+GCT_lang = {
+    \"prev\"                  : \"" . tr('pagination_left') . "\",
+    \"next\"                  : \"" . tr('pagination_right') . "\",
+    \"number_of_caches\"      : \"" . tr('number_of_caches') . "\",
+};
+");
+
+?>
```
andrixnet: well, removed it :-)

lib/search.ovl.inc.php
```diff
@@ -155,7 +155,7 @@
         }
 
         $nr = 1;
-        $s = dbcSearch->simpleQuery(
+        $s = $dbcSearch->simpleQuery(
```
andrixnet: Either it wasn't used for many months, or the few users that tried it (and failed) didn't bother to report it...

andrixnet: I know nothing about this format or what devices make use of it.

xml/ocxml11.php
```diff
@@ -365,7 +365,7 @@ function outputXmlFile($sessionid, $filenr, $bXmlDecl, $bOcXmlTag, $bDocType, $z
             FROM `cache_attrib` WHERE `language`='pl'");
         fwrite($f, $t1 . '<attrlist>' . "\n");
         while ($r = XDb::xFetchArray($rs)) {
-            fwrite($f, $t2 . '<attr id="' . $r['id'] . '" icon_large="' . xmlentities($absolute_server_URI . $r['icon_large']) . '" icon_no="' . xmlentities($absolute_server_URI . $r['icon_no']) . '" icon_undef="' . xmlentities($absolute_server_URI . $r['icon_undef']) . '">' . xmlcdata($r['text_long']) . '</attr>' . "\n");
```
andrixnet: Yeah, but still, though not included here, we have a new xmlentities function in export.inc.php.
I prefer to keep this rename because it is a local reimplementation of an already existing library (albeit not included here) function.

lib/languages/pl.php
andrixnet: done

PR #973, opened 2017-04-11
 	Fix conversion errors, improve performance

miklobit:
improve performance (DB import with low-end server)
 - limit data size per single INSERT
 - split schema import ( import data without triggers active )
convert binary data to hex 
remove all truncated data fields  (to get  reliable test data)

### DIFF
--- a/docs/vm/db_export.php
+++ b/docs/vm/db_export.php
@@ -17,11 +17,15 @@ $charset= 'utf8';
 // file with export pattern
 $exportPatternFile = "./db_export_ocpl_pattern.json";
 
-//if TRUE db structure will be attache at the begining of the result file
+//if TRUE db structure will be attached at the begining of the result file
 define('DB_STRUCT_DUMP', TRUE );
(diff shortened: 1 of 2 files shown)


andrixnet:
@miklobit what do you mean by "**low-end server**" ?

miklobit:
Low-end = VPS with resources as low as 1xCPU/1GB RAM.
Finally i was able, with such  resources, to import (almost) complete PL database  in abt. ~30 min.

andrixnet:
How many caches are there on OCPL (total), and how many log entries (these being the tables with most rows) ? 

Note: OCRO is running on VPS with 1CPU; 0.5GB RAM (as comparison, but our cache count doesn't compare :-) )

deg-pl:
@andrixnet why you don't use Arubacloud VPS Small - 1 EUR/month (+VAT) gives you 1 core, 1GB RAM and 20GB SSD storage! https://www.arubacloud.com/vps/virtual-private-server-range.aspx

Issue #988, opened 2017-04-26
search.php - can't select "wojewodztwo" after select Poland in combobox


kojoty:
![image](https://cloud.githubusercontent.com/assets/2961448/25452678/b2fe916a-2ac6-11e7-84a5-01cf83bf6c14.png)

It seems to be a regression after changes around GPX generation - I'll try to fix it now

PR #989, opened 2017-04-26
fix for JS error on form change (#988)

kojoty:
fast fix for JS errros at search.php at oc.pl (if size-1 and size-8 are not supported)


kojoty:
@andrixnet, I added you as a reviewer because this is fix for your changes: 
oc.pl doesn't support size-1 so this code generated JS exception and break the way of working search-page.

andrixnet:
I totally missed this one.

Issue #1003, opened 2017-05-06
Too much allocations in logGeokrety cronJob


kojoty:
There are many issues like
PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted
in apache error log last days.

This is caused at least by bug in GK queue processing algorithm.

kojoty:
I've made refactoring of GK-queue-processing and now it should work OK. It seems it's not enough, because we have some number of broken entries in queue, so probably procedure which add GK-logs to queue also needs fixes...

@andrixnet, @harrieklomp, @deg-pl 
I change the name of GK-processing script. Till now (at least at PL node) there was a line in cron config:
`#*/10 *  *   *   *   /srv/cron/cron-defs/do-wget-url util.sec/geokrety/logGeokretyCronJob.php logGeokretyCronJob.html`

now it should be rewrite to:
`#*/10 *  *   *   *   /srv/cron/cron-defs/do-wget-url util.sec/geokrety/processGeoKretyDbQueue.php processGeoKretyDbQueue.php`

New script (by default) is quiet and not produce any output - just tried to send GK-logs. All sent logs are removed from DB, all broken ones are still in DB for further debug (last-try column is set for them).

It is easy to debug this procedure by calling:
`wget --read-timeout=3600 --tries=1 http://opencaching.pl/util.sec/geokrety/processGeoKretyDbQueue.php?debug` from server - then output will contain detailed info (especially about broken GK-log entries)

andrixnet:
@kojoty : I presume the recommendation is also to activate this cron job by removing the `#` in front.
OCRO updated.

One question: util.sec/geokrety/geokrety.new.php remains valid?

kojoty:
> I presume the recommendation is also to activate this cron job by removing the # in front.
> OCRO updated.
@andrixnet, sure ;) and of course the 10' period can be changed as well.

> 
> One question: util.sec/geokrety/geokrety.new.php remains valid?
Yes, there are no changes in geokrety.new.

harrieklomp:
OCNL did not have such a line in cron. Did we miss something in the past?
We have added the line in cron.

deg-pl:
Crontab updated in OC UK and OC PL.

PR #1004, opened 2017-05-06
sql alter for GK processing fix (#1003)

kojoty:
We have an issue in GK log processing (which is called from cron). Now at oc.pl it hangs because of too many broken GK log records. To fix it I need a new column in gk_log table.

@harrieklomp, @andrixnet, @deg-pl - please apply db update and give me know here it is ready.
Thanks!

BTW. This is called from cron and shouldn't affect user activity so let's push it during the weekend.


deg-pl:
done @ OCPL and OCUK

harrieklomp:
Done on OCNL

kojoty:
@andrixnet, I'm waiting for your reaction. :)

andrixnet:
Done on OCRO. Sorry for my tardiness.

PR #1005, opened 2017-05-07
refactoring of GK logs DB queue processing (#1003)

kojoty:
I made wider refactoring than I was planned, but I believe now sending GK logs is a little bit simpler. 
Changes in context of #1003.


deg-pl:
@kojoty you've removed util.sec/geokrety/logGeokretyCronJob.php - should all nodes change anything in cron tabs before you'll merge it?

kojoty:
I've planned to add instructions for it, but first I would like to test it by hand at oc.pl server - maybe it will need some fast fixes - then all nodes should redefine cron rule for logGeokretyCronJob.

I'll add an empty file with old name for backward compatibility for short time of migration.

PR #1023, opened 2017-06-04
part of settings (config) moved to global context

kojoty:
It was supposed to be a small fix but every time I change something in common.inc.php I always have a  desire to remove everything from this file and I try to do it line by line :)


Utils/Text/TextConverter.php
```diff
@@ -0,0 +1,99 @@
...
+
+class TextConverter
+{
+
+    public static function addHyperlinkToURL($text){
+
+        $texti = mb_strtolower($text);
+        $retval = '';
+        $curpos = 0;
+        $starthttp = mb_strpos($texti, 'http://', $curpos);
```
deg-pl: What about https:// ?

kojoty: Good point - but I've not written this code - I've just moved it from common.inc.php - I even don't know if this is really used - it is used in 2 places in log edit but it could be a dead code as well.

PR #1028, opened 2017-06-07
general common.inc.php cleanup

kojoty:
I decided to go further with it - I hope now "everywhere loaded" code is simpler and more tidy


deg-pl:
Jedno słowo: SZACUN! :+1: 

(One word: RESPECT)

PR #1033, opened 2017-06-07
temporary moved more config vars to global context

kojoty:


andrixnet:
Some of these will get into the settingsGLUE I envisioned, until all configuration is rewritten into $config[] :-)

PR #1050, opened 2017-06-19
Refactoring powertrail to cacheSets

kojoty:
This is very early stub - I'll try to commit small changes, which shouldn't affected working code.

Why "cacheset"? We have a small discussion when I implemented powertrail/geopath/cacheset support for OKAPI, and there we decided that the best technical name for it is "cacheset" (BTW oc.de also have something similar called cacheSet).


Controllers/ViewCacheController.php
```diff
@@ -211,7 +211,7 @@ private function processOcTeamComments()
 
                 // add OC Team comment
                 GeoCacheDesc::UpdateAdminComment( $this->geocache, $_POST['rr_comment'], $this->loggedUser);
-                $this->view->redirect(Uri::getCurrentUri());
+                $this->view->redirect(Uri::getCurrentUri(true));
```
deg-pl: Thanks ;)

Issue #1053, opened 2017-06-20
Unwanted "micro" caches back on OCNL


harrieklomp:
After commit https://github.com/opencaching/opencaching-pl/pull/1049 it is possible again to make new caches on OCNL. In settings.inc.php is a rule that should prevent this, but somehow this is not working anymore.

![selectie_066](https://user-images.githubusercontent.com/1060518/27317221-3df7aa48-5585-11e7-88fc-8ba840a2d492.jpg)

![selectie_065](https://user-images.githubusercontent.com/1060518/27317252-5b9f8962-5585-11e7-88f7-c67205607449.jpg)

kojoty:
I think it is fixed now.

PR #1054, opened 2017-06-20
#1053: filter out forbidden cache sizes in newcache.php

kojoty:
@harrieklomp I don't think it had worked before my last changes - check the history of this file...
Anyway - now it should work for you.

BTW: I think we need to summarize the common data (as sizes, attributes, types etc.) @andrixnet started the work at opencaching.eu - I think I will need it in further works.


andrixnet:
@kojoty I will continue my work on wiki.opencaching.eu later this week

Issue #627, opened 2016-05-17
Unused files?


andrixnet:
I am investigating unused files in the source tree.
- [x] images/cacheatr/ - apparently referenced only in /tpl/stdstyle/cache-atr.tpl.php (which seems to be unused)
- [x] images/cachemaps - apparently not referenced
- [x] util/google-earth/\* and images/ge - apparently not referenced
- [x] images/header/ and util.sec/make_logo.php - apparently not referenced
- [x] images/thumbs - apparently not referenced
- [x] images/amnesius\* 
- [x] images/*.swf
- [x] images/diff-*
- [x] images/terr-*
- [x] images/lightbox/\* seem to be a duplicate. It is referenced, but /tpl/stdstyle/images/lightbox is actually used
- [x] various files under /images (especially third party logos)
- [x] lib/tinymce (not 4) seems to be a remnant from a previous tinymce instance ?
- [x] various .js files are spread around lib/ rather then being under lib/js/
- [x] SVG and graphic resource files under logbook/ should be moved as templates under docs/ and the files actually used to be published by each node via wiki, based on these templates
- [x] util/ and Utils/ should be merged. (no reason to have many utilities directories)
- [x] there are at least 3 versions on jQuery under tpl/stdstyle/js/ ; These should be unified to a single deployment.
- [x] tpl/stdstyle/images/log/ and tpl/stdstyle/images/log2 ; These should be merged or cleaned up
- [x] tpl/stdstyle/images/profile 
- [x] tpl/stdstyle/images/progressbar
- [x] tpl/stdstyle/images/thumb - at least some are unused (images with german text)
- [x] tpl/stdstyle/images/toltipsImages - should be added to more pages that need such icons (and directory renamed)

These are the ones I was able to identify today. 
If any/some/all of these are not used anymore, I suggest removal, they only add clutter.

kojoty:
I like removing files so I try to look at this in a meanwhile...

mzylowski:
util/google-earth/* and images/ge - apparently not referenced
We have reference to it from our wiki.
It is plugin to google-earth. If it still works it shouldn't be removed.

deg-pl:
@mzylowski I think we should move google-earth plugin to wiki OC PL and remove from github.

andrixnet:
Even if google-earth plugin involves no specific PHP and or JS code, (ie only KML file), I disagree. 
Plugin probably would need editing to make it specific to a certain node.

[...]

Issue #1074, opened 2017-06-30
Remove support for very old browsers


deg-pl:
In many places in OC PL code we have code for support very old browsers. For example in CSS:
`img                 { behavior: url(iepngfix/iepngfix.htc); }`
(It never worked because of invalid (relative to CSS) path!!!)
Or in start.tpl:
```
//detect browser:
    if ((navigator.appName == "Netscape" && parseInt(navigator.appVersion) >= 3) || parseInt(navigator.appVersion) >= 4) {
        rollOvers = 1;
    } else {
        if (navigator.appName == "Microsoft Internet Explorer" && parseInt(navigator.appVersion) >= 4) {
            rollOvers = 1;
        } else {
            rollOvers = 0;
        }
    }
```
Yeah! Support for IE <=3! (IE 4.0 was released in 1997!)

I think we should remove this code. Nobody uses such old versions, and if uses - should update immediately.

kojoty:
iepngfix is removed with this commit: https://github.com/opencaching/opencaching-pl/pull/1092/commits/d15779540fc8b00d07582987902747fcaf1ab218

PR #1092, opened 2017-07-19
627 unused files remove

kojoty:
in context of #627


kojoty:
additionally it affects #1074

PR #1094, opened 2017-07-19
#627: smiles.inc.php removed again

kojoty:
@deg-pl, @harrieklomp, @andrixnet - please note there is next sqlalter which fixes emoticons path in DB!


deg-pl:
Done on OC PL and OC UK. There are missing semicolons at the end of UPDATE lines.

harrieklomp:
Done on OCNL

andrixnet:
```
ERROR 1064 (42000) at line 6: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'UPDATE cache_desc SET `desc` = REPLACE(`desc`, 'lib/tinymce/plugins/emotions/ima' at line 2
```

Indeed, traced it down to semicolons missing.

andrixnet:
done on OCRO

PR #1100, opened 2017-07-22
current triggers from oc.pl db

kojoty:
triggers I've found in current oc.pl db - maybe not everyone is aware of it


deg-pl:
Great idea!
I think you can safe merge this PR without waiting for monday.

PR #1103, opened 2017-07-24
Next small OC improvements

deg-pl:


lib/languages/nl.php
```diff
@@ -201,7 +203,6 @@
     'new_password' => 'Nieuw wachtwoord',
     'new_topic' => 'New topic',
     'page_error' => 'Site error',
-    'contact' => 'Contact',
```
deg-pl: @harrieklomp - there was a duplicate of 'contact' entry in language file. I've removed one, but one was "Contact us", the second one - only "Contact". Could you check it please - which is correct.

lib/languages/ro.php
```diff
@@ -201,7 +203,6 @@
     'new_password' => 'Parolă nouă',
     'new_topic' => 'Subiect nou',
     'page_error' => 'Eroare site',
-    'contact' => 'Contact',
```
deg-pl: @andrixnet there was a duplicate of 'contact' entry in language file. I've removed one, but one was "Contactează-ne", the second one - only "Contact". Could you check it please - which is correct.

lib/languages/nl.php
harrieklomp: I have corrected it to "Contact".

deg-pl: Thanks!

PR #1104, opened 2017-07-26
powertrail_progress table + new triggers on cache_logs

kojoty:
@deg-pl, @harrieklomp, @andrixnet - please note:

This PR contains sqlalter with definition of the new table (powertrail_progress) and triggers + procedures used on cache_log table changes.

DB triggers are pure evil for code maintenance but I can't see any simpler method to be sure that every log is counted...


deg-pl:
@kojoty - VERY IMPORTANT!

On OC PL done without any problems (BUT...)
OC UK (MariaDB 10.1.25) generates error:

> #1235 - This version of MariaDB doesn't yet support 'multiple triggers with the same action time and event for one table'

On my devel (MariaDB 10.2.7) done without errors.
Now we have two triggers for cache_log: AFTER UPDATE (cl_update and yours - new) and AFTER INSERT (cl_insert and yours - new). Is this correct?
Or we should drop triggers cl_insert and cl_update and recalculate stats?

It is not a problem to update MariaDB to 10.2.x on OC UK, but I'm shure that @andrixnet (or @harrieklomp  - I don't remember who of them has MariaDB 5.x) will have bigger problem.

harrieklomp:
Done on OCNL and no errors. Even without any massages how many items that have changed.

deg-pl:
@harrieklomp have you "cl_update" and "cl_insert" triggers in DB?

harrieklomp:
I believe yes. I am not that technical to provide you with technical info so i took a screenshot via phpmyadmin.

selectie_090
I hope this answers your question?

deg-pl:
@harrieklomp - yes. Thanks.
PS: You have a very old version of phpmyadmin :)

kojoty:
@deg-pl I wanted to save an hour and leave those cl_* triggers behind - It seems they are used only for one script in stats - but now I see I need to reimplement and integrate it.  I tried to implement my triggers in a way which allow easy extend with procedures for another counters, so it shouldn't be difficult.

deg-pl:
SqlAlter done on OC UK (after upgrade MariaDB to 10.2.7)

andrixnet:
```
ERROR 1235 (42000) at line 134: This version of MariaDB doesn't yet support 'multiple triggers with the same action time and event for one table'
```
I am unable to upgrade MariaDB major version at this time due to hosting constraints.

andrixnet:
Code that does not work:
```
DROP TRIGGER IF EXISTS cache_logs_insert;;

CREATE TRIGGER cache_logs_insert AFTER INSERT ON `cache_logs`
    FOR EACH ROW begin
        IF NEW.`deleted`=0 THEN
           -- new, not-deleted log is now added
           CALL inc_logs_stats(NEW.`type`, NEW.`user_id`, NEW.`cache_id`);
        END IF;
    END;;
```

and 

```
DROP TRIGGER IF EXISTS cache_logs_update;;

CREATE TRIGGER cache_logs_update AFTER UPDATE ON `cache_logs`
    FOR EACH ROW begin
        IF OLD.`deleted`=0 AND NEW.`deleted`=0 THEN
            -- update of active log
            IF OLD.`type` <> NEW.`type` THEN
                CALL dec_logs_stats(OLD.`type`, OLD.`user_id`, OLD.`cache_id`);
                CALL inc_logs_stats(NEW.`type`, NEW.`user_id`, NEW.`cache_id`);
            END IF;

        ELSEIF OLD.`deleted`=1 AND NEW.`deleted`=0 THEN
            -- log UNDELETE
            CALL inc_logs_stats(NEW.`type`, NEW.`user_id`, NEW.`cache_id`);

        ELSEIF OLD.`deleted`=0 AND NEW.`deleted`=1 THEN
            -- log DELETE
            CALL dec_logs_stats(OLD.`type`, OLD.`user_id`, OLD.`cache_id`);

        ELSE -- OLD.deleted=1 AND NEW.deleted=1
            -- do NOTHING - update of removed log without status change
            CALL nop();
        END IF;
    END;;
```

andrixnet:
This one works:
```
DROP TRIGGER IF EXISTS cache_logs_delete;;

CREATE TRIGGER cache_logs_delete AFTER DELETE ON `cache_logs`
    FOR EACH ROW begin
        IF OLD.`deleted`=0 THEN
           -- not-deleted log is now removed
           CALL dec_logs_stats(OLD.`type`, OLD.`user_id`, OLD.`cache_id`);
        END IF;
    END;;
```

deg-pl:
@andrixnet - OC UK uses VPS from ArubaCloud (https://www.arubacloud.com/) for 1EUR/month! VPS gives freedom choosing software. Think about it ;)

andrixnet:
Actually the 1EUR/month VPS is restricted to several predefined guest images (none of which suit me), according to ArubaCloud documentation. I've researched it a few months ago.

I could eventually upgrade to MariaDB 10.0.x but only after a complete distro reinstall @ new version.

Question: can't trigger action code be merged into existing trigger?

deg-pl:
I think these triggers will run only on MariaDB 10.2.x. I've unsuccesfully tried on 10.1.25.

@kojoty What is your opinion about merging triggers?

kojoty:
I'm going to come back to "triggers" soon (this is also part of my big-slowly-geopatch-refactoring). Right now I'm working on different part of code.

Let me understand: older maria DB doesn't support two different triggers on on table right? So to make it work I need to merge all triggers (for given table) to one - right?

deg-pl:
Yes. Exactly!

andrixnet:
Does not support multiple triggers on the same table for the same action.

PR #1111, opened 2017-08-03
css loaded by JS in pagination chunk

kojoty:
css is loaded by JS to page header to not break HTML rules...


kojoty:
@deg-pl - ready!

deg-pl:
WOW! Thanks!

PR #1117, opened 2017-08-08
User preferences util implementation

kojoty:
This is separate util class which I would like to use in many places for fast save/read user preferences.

I understand 'user-preferences' as settings of UI which we would like to store for the user (like map filter settings or decisions which section of the page should be visible etc). 

User-preferences are not a User-settings (like a decision if user wants to be a OC guide in user profile) which needs to be stored separately, because we need for example search all users which are guides. User-preferences are in generally for minor decisions about UI etc.

Currently such preferences are stored in separate table (check map_settins, mapv2_settings db tables), where every setting is separate column - what - I think - is not a useful solution.

User preferences are stored in DB as simple JSON and can be accessed by the 'key'='name of set of user preferences'. 

For every user-preferences set of data (identified by key) should be defined simple class with default values (derived from `UserPreferencesBaseData` class).

- to SAVE set of user preferences (from some view/page) just implement ajax controller which needs to just call: `UserPreferences::savePreferencesJson(key, json)` method.

- to READ set of user preferences just call `UserPreferences::getUserPrefsByKey(key)` and then get final JSON `getJsonValues` method on returned object to get final JSON.

I will deliver example of usage of this code soon.

@deg-pl, @andrixnet, @harrieklomp - this PR contains sqlalter


harrieklomp:
Done on OCNL

deg-pl:
Done on OC PL and OC UK.

andrixnet:
done on OCRO

PR #1120, opened 2017-08-08
fix log error: Undefined offset: 0 in /srv/ocpl/powerTrail/ajaxGetCom…

kojoty:
…ments.php on line 69


kojoty:
BTW: this is only workaround on bug in powertrail code which allow to save comments with type without verification - so there is for example records in powertrail_comments with type=0 - now I just skip such comments.

Issue #730, opened 2016-08-15
[Security] Vulnerability by $_COOKIE through unzerialize()


Rotzbua:
same as: https://github.com/opencaching/okapi/issues/438
- Object Injection
- https://www.owasp.org/index.php/PHP_Object_Injection
- affects all OC versions
- replace serialize by json_decode/encode -> then also usable by javascript

Found at:
lib/cookie.class.php
mobile/lib/cookie.class.php
`$this->values = @unserialize($decoded);`

wrygiel:
Alternative: I've done some reading and I believe that in PHP 7 you can also call `unserialize($decoded, array('allowed_classes' => false))`.

deg-pl:
We have also similar issue #1020

kojoty:
@wrygiel, it seems to be a real problem. What do you think should be done here - I think we should fix ASAP.

wrygiel:
1. Easy, but a bit less secure - copy the [solution from OKAPI](https://github.com/opencaching/okapi/commit/d148501e24eeacdb115235a455d35219e0c446fb).
2. A bit more difficult, but more secure - copy the [solution from OCDE](https://github.com/OpencachingDeutschland/oc-server3/pull/364/files).

kojoty:
everything seems to work OK, so I guess we can close it.

PR #1123, opened 2017-08-09
serialize/unserialize in cookie cookie class (security issue)

kojoty:
in context of #730

### DIFF
--- a/lib/cookie.class.php
+++ b/lib/cookie.class.php
@@ -15,11 +15,14 @@ class cookie
         global $opt;
 
         if (isset($_COOKIE[$opt['cookie']['name'] . 'data'])) {
-            //get the cookievars-array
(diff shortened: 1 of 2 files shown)


kojoty:
This PR fixes the security issue, but if I understand well the side issue is that all our cookies with settings need to be reset...

@wrygiel, can you confirm that this is enough to fix it - thanks.

wrygiel:
Yes, this change will invalidate all current user session. Everyone will need to sign-in again. If some other variables are stored in this cookie, then they will also be lost.

BTW - this also needs to be change in OKAPI (in unison), because OKAPI currently expects OCPL's cookie to be serialized (not JSON-encoded).

wrygiel:
I think you have write access to OKAPI, so simply commit both changes one after another.

PR #1170, opened 2017-09-01
remove duplicate old OKAPI files; fixes #1162

### LINKED ISSUE #1162: Duplicate OKAPI files
There are lots of duplicate files in the "okapi" directory tree, which should have been deleted after renaming.

[This one](https://github.com/opencaching/opencaching-pl/tree/master/okapi/locale) makes problems on case-insensitive file systems; I will create a PR to remove the outdated lowercase version.

In "okapi/services" all files are duplicate, e.g services/caches/geocaches.php = services/caches/geocaches/WebService.php

following5:
@wrygiel 
Note that by duplication and deleting, the Git histories of these files are lost. But as it's only a deployment history that may be neglectable.


following5:
is included in #1186

PR #1175, opened 2017-09-05
OC Team menu refactoring part 1

deg-pl:


tpl/stdstyle/admin/reports_list.tpl.php
```diff
@@ -0,0 +1,54 @@
...
+        <input type="text" placeholder="{{waypointname_label}}" name="reportWp" value="<?=$_SESSION['reportWp']?>" class="form-control">
+        
+      </td>
+      <td><select name="reportType" class="form-control"><?=$view->typeSelect?></select></td>
+      <td><select name="reportStatus" class="form-control"><?=$view->statusSelect?></select></td>
+      <td><select name="reportUser" class="form-control"><?=$view->userSelect?></select></td>
+    </tr>
+<?php foreach ($view->reports as $report) { ?>
+    <tr>
+      <td><a href="/admin_reports.php?action=showreport&id=<?=$report->getId()?>" class="links"><?=$report->getId()?></a></td>
```
kojoty: Am I right that ampersands (&) needs to be encoded (&amp;) inside HTML attributes?

deg-pl: Will be fixed. Thanks

Issue #874, opened 2017-02-24
Make location of ocpl-update.sh configurable


andrixnet:
post-commit.php contains a hardcoded reference to /var/www/ocpl-update.sh as the update script for the site. 

This works fine when the site is hoted on a Linux server where root access is available to the OC admin.
Unfortunately the webhosting of OCNA is through a hosting company. Client user does not have permissions to write in /var/www

Also, in principle, having such a hardcoded reference is bad practice.

andrixnet:
Fixed in https://github.com/opencaching/opencaching-pl/pull/1192

mzylowski:
@andrixnet Issue will be closed automatically if you prepare right commit message (see: https://help.github.com/articles/closing-issues-using-keywords/).
If you want closing issues manually, please do it after merging fix to master, not when PR is submitting ;]

Issue #290, opened 2016-03-04
Site settings cleanup


kojoty:
@andrixnet:

> I want to tackle a serious improvement in the website settings section
> and I would like to ask for advice first.
> 
> Currently there is settingsDefault.inc.php which lists a lot of defaults.
> There is settings.inc.php which is individual to each node and overrides
> the defaults above.
> 
> -----------------
> What I intend is a cleanup and reorganization in a much better fashion
> all the existing settings, which most likely involves changing many
> existing variables (that are currently defined very chaotically) to a
> unique convention based on the $config array and introducing some new ones.
> 
> In order not to adversly affect the currently running production nodes,
> I believe the best option is the following:
> - keep the current settings.inc.php as it is, until all work is finished
> (may take a while)
> - add a glue file that includes settings.inc.php and is included in the
> code instead of settings.inc.php. The purpose of this file would be to
> override the new definitions in settingsDefault.inc.php using the old
> variable names from settings.inc.php.
> - change the settings-example.inc.php to the new definitions
> - make all the changes sequentially based on thematic sections. (easier
> to track, test and manage)
> 
> While the work is in progress, partial implementation updates can be
> deployed this way and the old variables in settings.inc.php are
> performing their purpose such that production nodes work properly.
> 
> When all changes have been migrated, all node server admins will be
> instructed to take the new settings-example.inc.php and recreate their
> settings.inc.php.
> When all node server admins have confirmed that these changes have been
> implemented, the glue file can be eliminated.
>

kojoty:
I think that cleaning up the mess around the configuration is a good idea and we should do that! 

My comments: 
- It seems to be a bigger work in the core of the website, but we are planned the second huge refactoring around mysql_* ( #276 ) and I think we shouldn't do that in the same time
- I think we should use lib/objects/OcConfig class for access interface to the config settings for the rest of the code, and then:
   - all settings should be closed inside this class (maybe in separate files, but it should be closed inside),
   - access to the config settings should be through the methods of this class
 
From my point of view it can be done in a steps:
- implement the OcConfig class - to be sure that all current settings can be accessed through it
- change all the "old-way" style accesses to the config across the whole code (it also can be done in many steps)
- clean up the config files, add sections etc. (and ocClass which is the only reader of them)

There are a lot of work, but this seems to be quite easy to do...

andrixnet:
One point I whish to stress greately: 
[...]

PR #1192, opened 2017-09-09
Settings refactor (first steps)

andrixnet:
First step to aid node admins during the settings refactoring over time, by not requiring them to change settings.inc.php with each new intermediate commit. 

No admin intervention is needed at this time. Only at the end of the project.

Part of project: https://github.com/opencaching/opencaching-pl/projects/3 and issue: https://github.com/opencaching/opencaching-pl/issues/290

Additionally, fixes https://github.com/opencaching/opencaching-pl/issues/874

### DIFF
--- a/post-commit.php
+++ b/post-commit.php
@@ -1,5 +1,7 @@
 <?php
 
+require_once __DIR__ . '/lib/settingsGlue.inc.php';
+
 # This script is called by GitHub after each commit. All changes commited to
 # the master branch are immediately checked out at the production server.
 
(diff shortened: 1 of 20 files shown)


post-commit.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+require_once __DIR__ . '/lib/common.inc.php';
```
kojoty: @andrixnet, I'm not sure if this line is a good idea. Common.inc contains and include lot of code. If anyone commit change with a bug in that code which generates some exception auto-update will not work anymore without admin interaction at every our node (manual git pull) - now you can revert the change at github and everything is OK again...

I think that maybe better just load only necessary config?

andrixnet: Yes, you are right. I'll fix it.

andrixnet:
Updated.

PR #1193, opened 2017-09-11
remove obsolete OKAPI files; fixes recent errors with search_and_retrieve

following5:
Again old files were not cleaned up when deploying to OCPL (commit eacfe040). This is a _critical_ fix which cleans the mess up.


following5:
thanks

PR #1220, opened 2017-09-18
Reports refactoring part 1e

deg-pl:
@andrixnet @harrieklomp please - run sqlAlters/2017-09-19-reports-refactoring.sql
Verify also you have index on reports.status - should be added via OKAPI update. If not - run:
```ALTER TABLE `reports` ADD INDEX(`status`);```

@kojoty now DB is ready for PowerTrail reports. reports.object_type should be set to Report::OBJECT_POWERTRAIL (2), reports.PowerTrail_id should be set to reported PT ID and reports.cache_id should be null.

@wrygiel @following5 Reports table now supports reports of not only caches, but also of PowerTrails/CacheSets. Your new OKAPI method to manipulate reports should have "WHERE object_type=1" SQL clause to retrieve reports of caches.


harrieklomp:
Run both sqlalters on OCNL

PR #1222, opened 2017-09-19
spelling: entery/enery => entry

following5:


kojoty:
I guess we have much more such errors. Thanks for cleanup these ones!

PR #1224, opened 2017-09-19
fix log text html encoding for new logs

### LINKED ISSUE #1218: XSS vulnerability in OCPL log handling code
It is possible to inject XSS into OCPL logs. 
An example here: https://github.com/opencaching/okapi/issues/412#issuecomment-329806320

Not sure how much if further OKAPI filtering is necessary, or just OCPL website code.

following5:
* fixes #1218 security issue for all newly submitted logs
* prepares migration of old logs text (migrated logs will be set to `text_html`=2)
* fixes some minor log HTML encoding issues

### DIFF
--- a/Controllers/CacheAdoptionController.php
+++ b/Controllers/CacheAdoptionController.php
@@ -151,8 +151,8 @@ class CacheAdoptionController extends BaseController
         // put log into cache logs.
         $logMessage = tr('adopt_32');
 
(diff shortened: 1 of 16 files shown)


following5:
I have now [applied](https://github.com/opencaching/opencaching-pl/commit/4ee0d1f31ed200879792791d294b83b718c3b426) this fix also to Okapi.

PR #1229, opened 2017-09-23
small cleanup around css

kojoty:
I've removed unused css selectors and mad a small cleanup around css files. It shouldn't change anything in page view.


deg-pl:
WOOOOW!
@kojoty we need also...
body {font-size:????} (I know... it is not easy, but next ones are easy!)
and classes .icon16 and .icon32 (hundred uses in code but... no definition!)

PR #1235, opened 2017-09-28
fix html cache descriptions in GPX; fixes #1126

### LINKED ISSUE #1126: GPX files
Apologies if I'm not using this forum correctly.

I have a problem with GPX files - they do not contain much of the HTML that is present on the cache page, not even` <p>` and `</p>`, so paragraphs all run together. This applies to both GPX and GPX GC format.

Rgds, Andy

following5:


kojoty:
thanks as well!

PR #1236, opened 2017-09-28
Reports - user side

deg-pl:


deg-pl:
@harrieklomp - I forgot adding few strings indeed. Thanks for support. Now all polish strings are translated info English. All subjects in mails related to reports have two prefixes: [standard_prefix] [R#reportId], so I removed {reportid} from few strings.

@andrixnet @harrieklomp - most (or all?) translations of reports now are on repository. Most important are reports_user_* because they are user related. Check also my english translation please ;)

PR #1239, opened 2017-10-02
fixed XML sequence error in GPX files

following5:
This puts some GPX lines into the right oder, to get rid of this XML validation error (tested with https://www.freeformatter.com/xml-validator-xsd.html):

`Cvc-complex-type.2.4.a: Invalid Content Was Found Starting With Element 'gsak:Child_Flag'. One Of '{"http://www.gsak.net/xmlv1/5":Resolution}' Is Expected.`

See definition in http://www.gsak.net/xmlv1/5/gsak.xsd..

For search.gpx it won't really matter, it has a bigger issue with the `<attributes>`. But search.gpxgc otherwise produces clean XML.


kojoty:
Thanks!

PR #1240, opened 2017-10-02
Add the Opencaching GPX extension (dont't merge yet)

following5:
This is a preliminary implementation of the new [Opencaching GPX extension](https://github.com/opencaching/gpx-extension-v1). It is just for review / testing / discussion, so **please don't merge**! Final code should be ready in 1-2 weeks.

This adds an `<oc:cache>` section for each geoache in the GPX files. Any apps which do not know that will ignore it; it is fully backward compatible.

The `<oc:cache>` section contains OC-specific information which otherwise is missing in GPX or not included in a safely machine-readable form:

* OC cache types
* log password requirement
* waypoint codes of other platforms, e.g. GC code
* estimated time and distance for searching
* UUIDs of log entries (this links GPX log entries to Okapi)

More information like OC attributes and OC waypoint types will be added in the future. It takes some time to define them properly.

Two detail notes:

* `<oc:size>` looks redundant to `<groundspeak:container>`. However, this is only the case for native OCPL GPX. Okapi and OCDE convert the sizes "Very large" and "Nano" to Groundspeak-compatible sizes "Large" and Micro". (Putting "Very large" and "Nano" into `<groundspeak:container>` has the drawback that some applications will not recognize it, then the cache as no size at all.) To have the same OC-GPX output on all plattforms, `<oc:size>` is redundantly added for OCPL.

* Similar for `<oc:site_team_entry>`. It is redundant to the log type "OC Team Comment". However, Okapi and OCDE don't have this log type, but a "Team comment" flag which can be set for any type of log entry. Again, to have the same OC-GPX output on all plattforms, `<oc:site_team_entry>` is redundantly added for OCPL.

### DIFF
--- a/lib/format.gpx.inc.php
+++ b/lib/format.gpx.inc.php
@@ -8,7 +8,7 @@ $nodeDetect = substr($absolute_server_URI, - 3, 2);
 
 $gpxHead = '<?xml version="1.0" encoding="utf-8"?>
 <gpx xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-    xsi:schemaLocation="http://www.topografix.com/GPX/1/0 http://www.topografix.com/GPX/1/0/gpx.xsd http://www.groundspeak.com/cache/1/0/1 http://www.groundspeak.com/cache/1/0/1/cache.xsd http://www.gsak.net/xmlv1/5 http://www.gsak.net/xmlv1/5/gsak.xsd"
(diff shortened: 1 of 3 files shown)


lib/format.gpx.inc.php
```diff
@@ -129,7 +149,6 @@
 
 // These strings must not be translated
 // Ref: see http://forum.opencaching.pl/viewtopic.php?p=121737#p121737
-$gpxContainer[0] = 'Unknown';       // OC: UNDEFINED
```
following5: Afaik 'Unknown' size is not definied for Groundspeak GPX. All other OC GPX implementations fall back to 'Other' size here, and I think this is the best solution.

(Practically, this fallback will only be used if someone forgot to add a new OC cache size to format.gpx.inc.php. So it's not that important.)

following5:
What is the prefered method at OCPL to resolve conflicts in open PRs? Rebase or merge-master?

kojoty:
> What is the prefered method at OCPL to resolve conflicts in open PRs? Rebase or merge-master?

It's up to you - we have many "merge" commits and this it is the easiest way. If you want to rebase it's only better.

following5:
Okay, then I will merge master here. But I will wait for #1006 / #1252, because the OC GPX extension somewhat increases the GPX file size.

lib/format.gpx.inc.php
andrixnet: It existed previously, I kept it as such when I worked on file exports

following5:
Oops, I acidentally deleted the branch. Will open a new PR later.

Issue #1200, opened 2017-09-13
Remove (?) logentries* tables


deg-pl:
We have two tables in DB: logentries and logentries_types. logentries_types is not used anywhere, In code we have only one INSERT to  logentries in /Utils/Log/Log.php, and no other access to this table. Why we collect such data? In PL DB we have 16254594 records - 3.7GB of unused data!
Could anybody explain what is purpose of logentries table? Or is it obsolete and can be removed?

andrixnet:
From what I see in my database at OCRO, these tables seem to be related to logging the actions performed by some cron jobs. 
For example it says watchlist module sent an email notification to userX @ timestamp and some other details.

andrixnet:
IMO this could be pruned to some interval (1 year ago, 6 months ago or similar) considering the purpose being logging cron job activities. It would be nice to be able to configure this, of course. 
At OCRO I have entries since I started the site in 2014.

deg-pl:
The questions are:
- are we need this logs?
- should logs to be stored in database???

andrixnet:
I've checked OCDE code and these tables also exist there. 
IMO they might have been in the german code before OCPL forked many years ago. 
OCDE code concept was to store everything in the database (also many data definitions that are still plaguing us). _(this works fine if there is only one production installation such as OCDE has with multiple views, but different installations that by design want to be in sync, that's not a good idea.)_

The purpose is clear: logging cron activity. IMO it is a good idea. It also helps keep track of cron activities and aid in administrative and debugging actions. 

Having this information stored in the database? IMO it is debatable. Since the code already does it as is, I don't think it's a majori priority to change that. It could be changed into a plain text log file.

However, regardless of logging be into database or file, logs should be pruned/rotated. Accumulation of such large quantities over a length of time is a bad administrative idea. 

IMO I would leave it in the database with the following tweaks: 
- delete new cache notifications after several months. Let's say 6
- delete watchlist notifications after 1-2 months (AFAIK these seem to be the most entries)
- delete owner, maintenance, status change and reporting notifications after at least twice the time an inactive cache is held before automatic archival. (so at least 12 months)
- similar decision logic to be used for other types. 

These deletions to be the object of an existing or new cron job, to run let's say once a month?

deg-pl:
Ok. Let's create new code do truncate these "logs".
Periods should be configurable per node.
Additionally - we should "truncate" CACHE_ACCESS_LOGS

deg-pl:
[...]

PR #1243, opened 2017-10-02
add indexes to optimize logging table cleanup

following5:
Whith these additional indexes, #1200 can be implemented without a new cronjob: Table cleanup will be fast enough to do it on-the-fly, by the same functions that insert new entries.

I would write the cleanup code for the tables then in a second step.

(table docs for `email_user` were outdated, have fixed that)

### DIFF
--- a/docs/database/approval_status
+++ b/docs/database/approval_status
@@ -26,7 +26,8 @@ Create Table: CREATE TABLE `approval_status` (
   `user_id` int(11) NOT NULL,
   `status` int(11) NOT NULL,
   `date_approval` datetime DEFAULT NULL,
-  PRIMARY KEY (`cache_id`)
+  PRIMARY KEY (`cache_id`),
+  KEY `date_approval` (`date_approval`)
 ) ENGINE=MyISAM DEFAULT CHARSET=utf8
 
 Changelog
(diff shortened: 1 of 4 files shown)


following5:
This kind of on-the-fly cleanup is done [here](https://github.com/opencaching/opencaching-pl/blob/master/search.php#L450) in search.php and [here](https://github.com/opencaching/opencaching-pl/blob/master/xml/ocxml11.php#L75) in ocxml11.php. It keeps things nicely at one place - data is deleted at the same place where it is created.

It has the drawback of consuming some disk space for the indexes. But I think that's only a small fraction of the whole data amount.

following5:
On the other hand - this is better done by someone who has access to the servers and can monitor the database, change settings etc. I will bail out from this.

deg-pl:
@harrieklomp @andrixnet - there is small sqlAlter to run.

@following5 always mention harrieklomp, andrixnet and me in PR which contains sqlAlter, please . We cannot track every code change, but we want have DB struct in sync on OC PL, RO, NL and UK.

Done on OC PL and OC UK.
WOW! index on logentries created 330 seconds!

following5:
ok then ...

harrieklomp:
Done on OCNL

andrixnet:
Done oc OCRO

PR #1246, opened 2017-10-03
added optional log tables cleanup; fixes #1200

### LINKED ISSUE #1200: Remove (?) logentries* tables

following5:
By default, nothing changes after deploying this commit. You then can enable the automatic cleanup by adding some settings (see lib/settingsDefault.inc.php, the last four lines). E.g.

`$config['logging_cleanup']['logentries'] = 12;`

will automatically remove entries from `logentries` table that are older than 12 months.

Of course the defaults in lib/settingsDefault.inc.php could be changed, so that there would be some basic cleanup at all OC sites (e.g. delete everything which is older than 24 months).

For **Opencaching.PL** I recommend to do a one-time manual cleanup **before** enabling automatic cleanup. This may take some minutes, but will overall speed things up:

```
delete from logentries where logtime < 'some-date';
delete from approval_status where date_approval < 'some-date';
delete from email_user where date_generated < 'some-date';
delete from CACHE_ACCESS_LOGS where event_date < 'some-date';
```
Instead of the last one, you may also run `util.sec/cleanup_access_logs.php`. It will remove all `CACHE_ACCESS_LOGS` that are older than 5 days.

Btw, I noticed that ALL database method calls in the code break PSR-2. The closing `)` bracket must go onto the start of the next line. I have kept it consistent here and added the database statement in the same non-PSR-2 style.

### DIFF
--- a/Utils/Log/CacheAccessLog.php
+++ b/Utils/Log/CacheAccessLog.php
@@ -11,12 +11,17 @@ use Utils\Database\XDb;
 
 class CacheAccessLog
 {
-
     const SOURCE_BROWSER = "B";
     const SOURCE_MOBILE = "M";
     const SOURCE_OKAPI = "O";
 
-    public static function logCacheAccess($cacheId, $userId, $event, $source){
+    public static function logCacheAccess($cacheId, $userId, $event, $source)
+    {
+        // There are many other places where rows are inserted into CACHE_ACCESS_LOG,
+        // but this probably is the most frequented one. So we do the cleanup just
+        // from here.
+        Log::cleanup('CACHE_ACCESS_LOGS');
+
         $accessLog = @$_SESSION['CACHE_ACCESS_LOG_VC_' . $userId];
         if ($accessLog === null) {
             $_SESSION['CACHE_ACCESS_LOG_VC_' . $userId] = array();
--- a/Utils/Log/Log.php
+++ b/Utils/Log/Log.php
@@ -6,16 +6,50 @@
 
 namespace Utils\Log;
 
+use Exception;
 use Utils\Database\XDb;
+
+$rootpath = __DIR__ . '/../../';
+require_once($rootpath . 'lib/settingsGlue.inc.php');
 
 class Log
 {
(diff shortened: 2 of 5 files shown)


deg-pl:
@following5 Thanks for your contribution.
I think better method is clean via dedicated cron job late at night. But do it daily (expiration period should be set in days, not months).
Your code will clean logs once per month during random USER activity. Let's relieve DB operations during user activity and make it dedicated at night.
Daily cleaning (and setting period in days) will be better in my opinion, because cleaning will be faster and the configuration will be more flexible.

@kojoty what do you think about it?

following5:
Oh, then I have misunderstood [this](https://github.com/opencaching/opencaching-pl/pull/1243#issuecomment-333763022). As I wrote in #1243, the new indexes are needed for on-the-fly cleanup. For a cronjob they are not needed.

This on-the-fly cleaning is already done [here](https://github.com/opencaching/opencaching-pl/blob/master/search.php#L450) in search.php and [here](https://github.com/opencaching/opencaching-pl/blob/master/xml/ocxml11.php#L75) in ocxml11.php and works fine. You won't notice any slowdown, deleting some indexed database rows is cheap.

following5:
(The advantage is that you dont't need to install cronjobs on all the four OC servers, cannot forget to install them, to disable them for some database maintenance or to reenable them afterwards. Noone can break the feature when fiddling around with crontab. But yes - it's unconventional.)

deg-pl:
@following5 We have 150 000+ viewcache.php views monthly. This is **150 000 UNNEEDED queries** DELETE FROM CACHE_ACCESS_LOGS (...)

following5:
okay ... I discard this

PR #1247, opened 2017-10-04
buffer large GPX downloads in file; use built-in Zip; fixes #1006

following5:
After deploying this change, everything should be like before.

To fix issue #1006, you then must add two settings. The first one goes into the Apache config:

```
<FilesMatch "\.gpx$">
    ForceType application/gpx+xml
    Header set Content-Disposition attachment
</FilesMatch>
```

If you have separate config files for http and https, you may need to add it twice. Then restart the webserver (e.g. `systemctl restart httpd.service`) and add this to lib/settings.inc.php:

`$config['downloadMemorylimit'] = 1024 * 1024 * 8;`

This will buffer all GPX data larger than 8 MB in a file, and then redirect to this file. (You may chose some other limit than 8 MB).

PR #1252, opened 2017-10-05
buffer large GPX downloads in file; use built-in Zip; fixes #1006

### LINKED ISSUE #1006: Gpx export memory issue
from oc.pl forum:
> Post autor: Kipcior » poniedziałek, 8 maja 2017, 08:47
> Żeby nie zakładać nowego tematu. Dzisiaj mam problem z eksportem wyników wyszukiwania do gpx. Jak zawsze wybrałem kryteria: Polska, wielkopolskie, reszta bez wykluczeń. Wyrzuciło mi liste ponad 3800 skrzynek. Po kliknięciu eksport do gpx wszystkich wyników wyszukiwania mieliło mieliło by po jakims czasie pokazac info:
> Firefox nie może odnaleźć pliku https://opencaching.pl/ocplgpx9909973.g ... =max&zip=1.
> Próbowałem kilka razy, zmieniając kryteria. Zawsze ten sam rezultat.

error example in log: 
[Mon May 08 09:38:35.706903 2017] [:error] [pid 28018] [client 81.15.211.114:5293] PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 100708352 bytes) in /srv/ocpl/lib/search.gpxgc.inc.php on line 565, referer: https://opencaching.pl/search.php?searchto=searchbyname&showresult=1&expert=0&output=HTML&sort=bydistance&f_inactive=0&f_ignored=0&f_userfound=0&f_userowner=0&f_watched=0&f_geokret=0&country=PL&region=PL41&cachetype=1111111110&cache_attribs=&cache_attribs_not=&cachesize_1=1&cachesize_2=1&cachesize_3=1&cachesize_4=1&cachesize_5=1&cachesize_6=1&cachesize_7=1&cachesize_8=%7Bcachesize_8%7D&cachevote_1=-3&cachevote_2=3.000&cachenovote=1&cachedifficulty_1=1&cachedifficulty_2=5&cacheterrain_1=1&cacheterrain_2=5&cacherating=0&cachename=%25&gpxLogLimit=100&cachename=

following5:
(followup to #1247)

After deploying this change, everything should be as before. To fix issue #1006, you then must add this configuration setting:

`$config['downloadMemorylimit'] = 1024 * 1024 * 8;`

This will buffer all GPX data larger than 8 MB in a file, and then redirect to this file. (You may chose some other limit than 8 MB).

By default, those temporary GPX files will be located in a "gpx" subdirectory below the image upload directory. You _may_ move this gpx directory to somewhere else by adding the two settings

* `$config['downloadPath']`
* `$config['downloadUrl']`

See settingsDefault.inc.php and settings-examples.inc.php for more instructions on that.

I have also rewritten the Zipping, which currently is disabled because it was too slow for large downloads. It may worth to give it a new try (later).

### DIFF
--- a/lib/search.gpxgc.inc.php
+++ b/lib/search.gpxgc.inc.php
@@ -1,16 +1,30 @@
 <?php
 
-ob_start();
-
 use Utils\Database\XDb;
 use Utils\Database\OcDb;
 use lib\Objects\GeoCache\GeoCacheCommons;
+
 global $content, $bUseZip, $usr, $hide_coords, $dbcSearch, $queryFilter;
 
 require_once($rootpath . 'lib/format.gpx.inc.php');
 require_once($rootpath . 'lib/calculation.inc.php');
 
 set_time_limit(1800);
+
+// cleanup old temporary files
+
+$downloadPath = $picdir . '/';
+if (is_dir($downloadPath . 'gpx')) {
+    $olddirs = glob($downloadPath . 'gpx/*', GLOB_ONLYDIR);
+    foreach ($olddirs as $dir) {
+        if (substr($dir, 0, 1) != '.') {
+            $stat = stat($dir);
+            if (time() - $stat['atime'] > 60*3) {   // delete after 30  minutes
+               exec('rm -rf ' . $dir);
+            }
+        }
+    }
+}
 
 function getPictures($cacheid, $picturescount)
 {
@@ -197,11 +211,10 @@ if ($usr || ! $hide_coords) {
 
     // Implementation ready, but do not use ZIP for now.
     $bUseZip = false;
-    if ($bUseZip == true) {
-        $content = '';
-        require_once ($rootpath . 'lib/phpzip/ss_zip.class.php');
-        $phpzip = new ss_zip('', 6);
-    }
+
+    $outfile = false;
+    $dirCreated = false;
+    $downloadMemoryLimit = ($config['downloadMemorylimit'] > 0 ? $config['downloadMemorylimit'] : PHP_INT_MAX);
 
     $children = '';
     $time = date($gpxTimeFormat, time());
@@ -217,8 +230,7 @@ if ($usr || ! $hide_coords) {
             $children = "(HasChildren)";
         }
     }
-    $gpxHead = str_replace('{wpchildren}', $children, $gpxHead);
-    echo $gpxHead;
+    $gpxData = str_replace('{wpchildren}', $children, $gpxHead);
 
     $stmt = XDb::xSql(
         'SELECT `gpxcontent`.`cache_id` `cacheid`, `gpxcontent`.`longitude` `longitude`,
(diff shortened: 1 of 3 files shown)


kojoty:
@following5, the same here - can we merge this PR? 
is it necessary to create "download" ($config['downloadPath']) dir manually? I mean if it needs any action from the server admin or it rather should work well as-is?

following5:
It is ready to merge. 

The download directory will be created automatically with the first download that is larger than the `downloadMemorylimit` setting.

The `downloadPath` / `downloadUrl` settings work like `$picdir` and `$picurl`, so they should look similar.

Please let me know when you have configured `downloadMemorylimit`, so that I can test that!

kojoty:
@deg-pl, can you set in OCPL config `downloadMemorylimit` mentioned by following5? Thanks.

deg-pl:
This is not good solution.
I've added downloadMemorylimit and tried to export big (28765 caches) GPX. Server stoped to respond for many minutes. After over 10 minutes I restarted apache. There was no error in error.log.

andrixnet:
What loadavg and what memory usage?

deg-pl:
LA was 1.3, memory - I don't remember, and I don't want to repeat test ;)

following5:
See https://github.com/opencaching/opencaching-pl/issues/1006#issuecomment-335944227. The server was NOT blocked.

following5:
... and the load was not changed by this commit.

deg-pl:
@following5 Ok, ok. You are doing great work! We should only prevent server for processing data for such long time. Maybe your cache limit is solution? Maybe limit should be even lower?

following5:
Have there been any problems with this long processing time? 30 minutes are explicitly allowed in search.gpxgc.php. This was implemeted [here](https://github.com/opencaching/opencaching-pl/commit/2227302c5c4cd2b148941e47da4deeb4416bd9b3#diff-2775af53c16b7a4cacb07eef8f354447), eight years ago.

deg-pl:
I think it has no sense with 30 minut max processing time. Nobody will wait so long.
@kojoty @andrixnet @harrieklomp What do you think about? What is good max processing time in your opinion?

following5:
I think it's much better to limit the number of caches than the processing time. The user then can relie that the request is fulfilled.

The processing speed is roughly 1000 caches per minute at Opencaching.pl. So if you say: It should normally not take longer than 5 minutes until the download starts, then the limit should be 5000 caches.

andrixnet:
There are several more considerations about this. 
The above assumptions are for a single user doing such an export at a time. 
Of course, if any other user happens to request another export while the first user's request is still being processed,
1. the server will overload on CPU and possibly RAM usage, 
2. processing time will increase accordingly, that is twofold or more. 

Furthermore: IMO 1 minute is already a **very very** long time to wait for a download to ... start, let alone 5 minutes. While it is the simplest technical solution by far, it fails by server overload or meeting reasonable user interraction expectations. 

---------------------------------------
Consequently I propose a slightly different approach, though technically more complex to implement, it can address all of the above in a much more efficient and elegant (towards the user) manner: 

A GPX export is basically a query on caches table with X caches as result (possibly further filtering with attributes), then a wider join with attributes, logs and waypoints. From MySQL point of view, such a query, even with several thousands of results, should return fairly quickly.

1. run the search query with count(*) to obtain the total number of results (caches) to process
2. take the total number of results and use it to:
2.1. split it into sections of Y caches to process at a time (this will prevent RAM usage overload)
2.2. have the ability to display a progress bar (percentage) of processed vs. total
3. after initiating download and obtaining total number of results, bring user to a page (or show an overlay) to inform the user about the GPX download progress
4. use AJAX to interract with the server to update the progress bar
5. reduce CPU priority (adjust niceness) of GPX export code section (allow more concurrent downloads)
6. deliver each section of GPX data over the network, once processed, to avoid RAM overload (buffering)
7. display a completion confirmation after all export has been processed.

These steps will offer the user a comprehensive interface that will avoid any frustration or abandonment due to perceived server lack of reponse, will inform the user about the progress of the operation, will allow more users to concurrently perform GPX exports, will prevent excessive CPU and RAM usage that might otherwise affect other user's access to the website and overall better end-user experience.

This works for both plain text export and zipped data. 
Of course, it can be extended into a common export interface from end-user point of view for all formats (OCPL generated at least)

andrixnet:
I just realized this is a PR, after I finished writing. 
Put my proposal in #1273.

Regarding the suggested limit in this PR: 5 minutes to be an absolute maximum, though I strongly suggest 1minute. (30 or 60 seconds is the connection timeout that a user might be reasonable expected to wait without too much frustration)

following5:
> IMO 1 minute is already a very very long time to wait for a download to ... start, let alone 5 minutes. While it is the simplest technical solution by far, it fails by server overload or meeting reasonable user interraction expectations.

Well... it has been eight years this way, and did anyone complain about download time or server load? ;-)

Of course it could be done better. But who will implement that?

deg-pl:
I think nobody will wait longer than 2-3 minutes. Maybe 5 minutes with progress bar.

PR #1256, opened 2017-10-08
translated German source code comments to English

following5:


kojoty:
@following5, thanks for that :)

Issue #531, opened 2016-04-16
GPX/other formats export _own_ unpublished caches


andrixnet:
This is a feature request of the GPX / other formats export module.

As owner of a cache that it is not yet published, but it's state is "to be published on [date]", I desire to export GPX with this cache (from cache page). Furthermore, I desire to add this (and other caches in the same situation) to clipboard. From the clipboard I can print them, but currently I cannot export them. 
As owner, I need all these caches exported to GPX as well. This refers to my own caches, of course.

Currently only published caches are exported.

deg-pl:
I think now is OK!
Owner CAN export his own unpublished cache via viewcache -> Download as file -> GPX

harrieklomp:
I think it is not okay yet ;-)
The downloaded GPX has only the site info. Please look at the screenshot.
![selectie_162](https://cloud.githubusercontent.com/assets/1060518/14600792/d3e92310-055e-11e6-800d-3044bef49865.png)

deg-pl:
Now I understand ;)

andrixnet:
Individually, but how about using the clipboard?

andrixnet:
@deg-pl @kojoty @mzylowski 
As I've been working on the export modules character encoding issues, I am willing to tackle this. 
In your opinion, what would be the best way to make this happen? 
- make all own unpublished caches exportable ?
- make special UI and exception to such export ?

This is not for search page results. Only for exporting files (which uses the search subsystem)
My main goal is to be able to export single caches and caches added to a list (clipboard).

following5:
Fixed by 1a8e6e5, issue can be closed

PR #1261, opened 2017-10-10
show unpublished own caches in search result; fixes #186

### LINKED ISSUE #186: search page to return also unpublished caches owned by myself
```
Hi, 

The search page explicitely refuses (line 469) caches that are in status 4, 5 
and (if not admin) 6.

Caches with status 4 and 5 are waiting to be published. Of course one must not 
see caches not yet published by other users. 
However, I own some yet unpublished caches and I want to export them into GPX 
file. However because of this limitation, I cannot. 

Preparing events and having caches to publish at a certain time and having to 
prepare GPX file in advance, well, you can understand my predicament. 

It's still too complex for me to attempt alone, and this should be done with 
care and elegantly.

Thank you. 

```

Original issue reported on code.google.com by `andrixnet` on 23 Apr 2015 at 10:03

following5:
To avoid the effort of drawing lots of new icons, the name of unpublished caches is displayed in red color. This color also applies to blocked caches which are visible for admins.


following5:
This also fixes issue #531.

kojoty:
@following5, thanks for the next fix :)

PR #1266, opened 2017-10-11
Buffer large GPX downloads in file, 3rd PR

following5:
[fix for #1006]

I have disabled the session blocking here and limited the number of caches to 10000.  I think this is a reasonable limit, which 
* does not allocate too much disk space (~ 150 MB for 10000 caches)
* has some headroom e.g. for the OC GPX extension, which somewhat increases GPX size
* allows to enable Zipping without making the overall GPX generation too slow.

If this works, we could go on for futher fine tuning, optimizations, try to enable zipping, add user interface to split larger downloads in several 10000-blocks, and so on.


following5:
needs further discussion

PR #1271, opened 2017-10-13
fixed OKAPI GPX download; fixes #1270

### LINKED ISSUE #1270: OKAPI GPX download is broken
The OKAPI GPX download of search results no longer works. On the OC.PL website I get an empty page, and on my develsite the error message:

Fatal error: Class 'okapi\Facade' not found in /mnt/ocpl/lib/okapi_gpx.php on line 16

following5:
This is part 1 of the bugfix. Part 2 is aready deployed as 5577d29e.

### DIFF
--- a/okapi_settings.php
+++ b/okapi_settings.php
@@ -51,7 +51,7 @@ function get_okapi_settings()
         'IMAGE_MAX_UPLOAD_SIZE' => $config['limits']['image']['filesize'] * 1024 * 1024,
         'IMAGE_MAX_PIXEL_COUNT' => $config['limits']['image']['height'] * $config['limits']['image']['width'],
         'OC_NODE_ID' => $oc_nodeid,
-        'OC_COOKIE_NAME' => $config['cookie']['name'],
+        'OC_COOKIE_NAME' => $config['cookie']['name'].'_auth',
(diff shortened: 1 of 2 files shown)


okapi_settings.php
```diff
@@ -51,7 +51,7 @@ function get_okapi_settings()
         'IMAGE_MAX_UPLOAD_SIZE' => $config['limits']['image']['filesize'] * 1024 * 1024,
         'IMAGE_MAX_PIXEL_COUNT' => $config['limits']['image']['height'] * $config['limits']['image']['width'],
         'OC_NODE_ID' => $oc_nodeid,
-        'OC_COOKIE_NAME' => $config['cookie']['name'],
+        'OC_COOKIE_NAME' => $config['cookie']['name'].'_auth',
```
kojoty: @following5, let me understand how OKAPI uses cookie?  Is it necessary for authorization? If yes - I'm afraid it's not enough, if not this change is not necessary...

following5: Okapi needs to know the session ID of the logged-in user. Then it queries the user ID from the `sys_session` table.

I have tested this change and it worked. :-)

following5: Please see 5577d29e, that is the corresponding Okapi-side change.

kojoty: Got it - now everything is clear - I've missed that OKAPI needs user sessionId here... Thanks for this fix - is it ready to merge?

following5: yep

Issue #204, opened 2015-09-13
Download own logs via GPX


harrieklomp:
There are people who like to download their own logs via a GPX. Is it possible to implement this in the code?
For example click in left menu on "My statistics" and click on "show" on the line "Number of found caches:". Could on this resultpage downloadable button(s) be added, like we have on the resultpage after a search via the searchpage?

wloczynutka:
Logs? oc do not allow downloading logs I think. We only porovide
downloading geocaches (including logs).

caches Found by specified user Can be done via search.php, (you can specify
finderid)

(
http://opencaching.pl/search.php?showresult=1&expert=0&output=HTML&sort=byname&searchbyfinder=&f_inactive=0&f_ignored=0&f_userfound=0&f_userowner=0&finderid=9067
)

Logged user can acces it from user profile.

2015-09-13 22:45 GMT+02:00 harrieklomp <notifications@github.com>:

> There are people who like to download their own logs via a GPX. Is it
> possible to implement this in the code?
> For example click in left menu on "My statistics" and click on "show" on
> the line "Number of found caches:". Could on this resultpage downloadable
> button(s) be added, like we have on the resultpage after a search via the
> searchpage?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/opencaching/opencaching-pl/issues/204>.
>



-- 
__________________________________
*Andrzej 'Łza' Woźniak*

*(+48) 798 835 592*
skype praca: *andrzej.wozniak.traveltech*
skype dom: *wloczynutka*

harrieklomp:
That own logs can not be downloaded on OC i know, that is why i ask if this can be implemented in the code.
Please look at http://www.opencaching.nl/my_logsFind.php?userid=3 or change the link to .pl and your own user ID. Could this page have his own "Download as GPX" button?

wloczynutka:
But you want download geocaches or only logs?

harrieklomp:
I only want to download my own logs.

wloczynutka:
At the moment we have no possibility to export pure log enteries at all. This is more than add button to mentioned page - You need to write code taking log-enteries from database, and exporting as xml / text / csv / whatever.

following5:
The GPX format does not allow to store logs without cache listings. A GPX file basically is a list of waypoints with coordinates and other properties, and logs have no coordinates. They just may be appended to waypoints.

You can download your logs via OKAPI, for whatever purpose you need them:
[...]

PR #1272, opened 2017-10-13
Add the "Include only my logs" option to OKAPI GPX download

following5:
This implements the "download my logs via GPX" function that @harrieklomp requested in #204.

### DIFF
--- a/lib/search.html.inc.php
+++ b/lib/search.html.inc.php
@@ -564,6 +564,20 @@ for ($i = 0; $i < $dbcSearch->rowCount($s); $i ++) {
     $caches_output .= $tmpline;
 }
 
+if ($usr !== false) {
+    // For the OKAPI GPX option "my logs". We may simplify and optimize this
+    // by adding a boolean "my logs" shortcut to the OKAPI GPX method.
+
+    $uuid_q = 'SELECT `uuid` FROM `user` WHERE `user_id` = :userid';
+    $params = [];
+    $params['userid']['value'] = (integer) $usr['userid'];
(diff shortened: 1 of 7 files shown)


tpl/stdstyle/js/okapiGpxFormatterWidget.pl.js
```diff
@@ -19,6 +19,7 @@
         paramLpcHeader: "Wpisy do logów:",
         paramLpc_0: "Nie dołączaj żadnych wpisów.",
         paramLpc_10: "Dołącz tylko 10 najświeższych wpisów.",
+        paramLpc_mine: "Include only my logs.",  /* TODO: translate */
```
following5: Needs Polish translation. Increment the `VERSION = 3` in line 46 to `4` with the translation.

(You may push the translation direct via [my branch](https://github.com/following5/opencaching-pl/tree/gpx-mylogs) into this PR, if you know how that works. ;)

following5: This is how it looks:

okapigpx

lib/search.html.inc.php
```diff
@@ -564,6 +564,20 @@ function fHideColumn($nr, $set)
     $caches_output .= $tmpline;
 }
 
+if ($usr !== false) {
+    // For the OKAPI GPX option "my logs". We may simplify and optimize this
+    // by adding a boolean "my logs" shortcut to the OKAPI GPX method.
```
following5: I will implement this optimization first and then issue a new PR. makes this change much simpler.

Issue #1268, opened 2017-10-12
Invalid target redirect after logon on OCPL


rapotek:
Steps to reproduce:

1. Logout from the service
2. Go to https://opencaching.pl/myneighborhood.php - redirects correctly to https://opencaching.pl/login.php?target=myneighborhood.php
3. Successful logon redirects to https://opencaching.plmyneighborhood.php without a slash ('/') in the middle, which is resolved to bad address

kojoty:
I know the problem - I try to fix it soon.

kojoty:
I hope now it is fixed.

PR #1274, opened 2017-10-13
#1268 fix of post-login redirect target

kojoty:


kojoty:
I merge in Friday because this bug can affect many users

PR #1276, opened 2017-10-13
implemented the Opencaching GPX extension

following5:
[replacement for PR #1240, which I acidentally closed]

This adds the [Opencaching GPX extension](https://github.com/opencaching/gpx-extension-v1) to all downloaded GPX files. The GPX files then contain an `<oc:cache>` section for each geoache. Apps which do not know that will ignore it; it is fully backward compatible.

The `<oc:cache>` section contains OC-specific information which otherwise is missing in GPX or not included in a safely machine-readable form:

* OC cache types
* log password requirement
* waypoint codes of other platforms, e.g. GC code
* estimated time and distance for searching
* UUIDs of log entries (this links GPX log entries to Okapi)

More information like OC log types and waypoint types may be added in the future.

The size of GPX files will increase by about 5-10%. The time to wait for the download will only slightly increase, I guess by < 1%, because there are no new database queries.

### DIFF
--- a/lib/format.gpx.inc.php
+++ b/lib/format.gpx.inc.php
@@ -8,7 +8,7 @@ $nodeDetect = substr($absolute_server_URI, - 3, 2);
 
 $gpxHead = '<?xml version="1.0" encoding="utf-8"?>
 <gpx xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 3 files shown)


kojoty:
@following5 - I assume we can merge this PR now - right?

following5:
yes

following5:
Thanks! The OC GPX extension now is only on all up-to-date OC sites, native and Okapi.

This is a big thing. We have defined a new standard, which allows to promote Openaching features and to implement new features in a way which is available to _all_ app developers.

following5:
only -> online

Issue #958, opened 2017-04-02
$nodeDetect strategy is wrong


andrixnet:
```
// sitename and slogan iternational handling
$nodeDetect = substr($absolute_server_URI, - 3, 2);

```

This is based on the website URL, and tries to obtain the site language. 
This will fail for any TLD suffix that is not a 2 letter language code

deg-pl:
It is related with #1137

I think site language should be stored in config, not readed from URI

kojoty:
It seems this is not an issue now - am I wrong? If yes - please reopen.

Issue #1075, opened 2017-06-30
New, responsive site interface


deg-pl:
Our site hasn't responsive interface. On smartphones looks not good. In my opinion - we should convert (or make brand new) to modern, responsive interface. We should (among others):
- do a new menu. With much less positions and horizontal, responsive layout
- simplify interface! Now we have plenty of data and controls. **New users cannot use our complicated interface**! For example - viewcache has 3 different date - hidden, created and last modified. But... none of them is important for users! Important is publication date, but we don't have this data!
- conversion menus, different texts into simple icons. Simplification is good way.
- move less important data to less important places. For example - we have "API" and "RSS" on main horizontal menu, these are options for a very small percent of users.
- drop cachemap v2

Plenty of users choose PAID GC because of simply, modern and responsive interface. We have a prehistoric interface... :(

If we will have modern, responsive interface - we can drop very old and undeveloped "mobile" version.

(related with #1073 #837, #764, #779, #510, #154 and many others)

andrixnet:
I agree, proper display on mobile device browsers is a must. 
Also, the current mobile site is antique, from the days of WAP browsers I think. 

A well thought, more modern and responsive UI is very good. 
There are menu items that don't really belong where they are from usability point of view. 
Some places use (at least in part) tables for layout (very old web practice). 

This however does not mean remove functionality from the user.

I do agree that the multiple dates might be too much, but: 
- date hidden: this is important for both the owner and the searcher.
- date published: this is relevant to the lifecycle of the cache on the website
- date created (creation of listing) = relevant only administratively
- last modified = relevant both administratively and for the user. Could also be shown as an age counter (ie modified 2 months ago)

I would suggest reworking some of the graphical artwork as SVG. Modern browsers support SVG well as images <img> tag. 

Menu should not clutter the display, but at least on desktop the menu should always be shown. Hiding the menu introduces additional clicks and mouse travel in user interraction, apparently simplifies the visible UI, but in reality complicating matters for the end user.

However, I wouldn't take as a comparison reference the "new" UI of GC, which is just horrible. It may be responsive, but it is slow, blocky, too much simplification and very serious losses of functionality, with other useless, low interest or annoying stuff brought forth.

deg-pl:
@andrixnet 

> date hidden: this is important for both the owner and the searcher.
Why???

> date published: this is relevant to the lifecycle of the cache on the website
Yes. I think this is only important date.

> date created (creation of listing) = relevant only administratively
[...]

Issue #1137, opened 2017-08-15
Support for all TLD


deg-pl:
In OC PL in many places we have code like this:

`$nodeDetect = substr($absolute_server_URI, - 3, 2);
`

But it is incompatibile with many TLD, for example - opencaching.game or opencaching.cloud. I think we should support all allowed TLD.

andrixnet:
Agreed

kojoty:
It seems this is not an issue now - am I wrong? If yes - please reopen.

PR #1282, opened 2017-10-16
New layout

kojoty:
This PR contains the proposition of the new layout - this is not ready yet - this is only an early stage of works. This not affect current pages - it can be seen by test.php only. There is still lack of menu-handling code - I'm working on it now.

Like I wrote in emails - the main idea is to create the similar layout to current common/main.tpl.php but with bootstrap css. This is a step on a way for further responsive design. We will be able to migrate pages to the bootstrap-like styles.

This is little in context of  #1075. 

@deg-pl, @andrixnet, @harrieklomp: 
I need config change (before commit this PR):
please add to your settings.inc.php line: `$config['ocNode'] = "<your-node-id>";`
For example for OCPL it should be `$config['ocNode'] = "pl";` 'ro' for OCRO, and so on...

It is replacement of node detection by domain which was used before in many places in the code.

If you want to look at this there is also my proposition of the "new" template's subsystem. TestController doesn't use old templates and eval - there are two simple methods which allow to build view-template file. Template contains only pure php injections to html. All values used in template have to be injected by $view->var, translations are done by $tr() function - everything is (in my opinion simple, clear and useful) - the only disadvantages is that var injection needs a little more chars - <?=...?> instead of {} - but there is huge advantage of this solution - there is no need of any parsing and rewriting of template file - this is just php file which is just included in the right place and time. Please take a look.

### DIFF
--- a/Utils/View/View.php
+++ b/Utils/View/View.php
@@ -4,8 +4,7 @@ namespace Utils\View;
 use Utils\DateTime\Year;
 use lib\Objects\ApplicationContainer;
 use Utils\Debug\Debug;
-use Utils\I18n\I18n;
-use Utils\Uri\Uri;
+use Controllers\PageLayout\MainLayoutController;
 
 class View {
 
@@ -220,7 +219,9 @@ class View {
     }
 
     /**
-     * TODO
+     * Wrapper for obsolate template system.
+     * Use display() instead!
+     *
      */
     public function buildView()
     {
@@ -228,15 +229,15 @@ class View {
     }
 
     /**
-     * TODO
-     * @param unknown $layoutTemplate
+     * Build template and display page.
+     * @param string|cont $layoutTemplate - base template to use
      */
     public function display($layoutTemplate=null)
     {
 
         if(is_null($layoutTemplate)){
-            $layoutTemplate = 'common/mainLayout';
-            $this->initMainLayout();
+            $layoutTemplate = MainLayoutController::MAIN_TEMPLATE;
+            MainLayoutController::init(); // init vars for main-layout
         }
 
         $this->_callTemplate($layoutTemplate);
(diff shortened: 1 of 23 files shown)


harrieklomp:
Add new lines on settings.inc.php on OCNL

Utils/View/View.php
```diff
@@ -184,23 +202,70 @@ public function addLocalCss($css_url){
...
+    {
         //TODO: refactoring needed but this is still this way
         tpl_set_tplname($tplName);
+        $this->_template = $tplName;
     }
 
-    public function buildView(){
-        //TODO: refactoring needed but this is still this way
+    /**
+     * Wrapper for obsolate template system.
```
deg-pl: 3 szyszki. Choć za komentarz mogą być 2. (obsolete)

index.php
```diff
@@ -41,8 +41,7 @@
 include ($dynstylepath . "totalstats.inc.php");
 
 // diffrent oc server handling: display proper info depend on server running the code
-$nodeDetect = substr($absolute_server_URI, - 3, 2);
```
deg-pl: :+1:

lib/settingsDefault.inc.php
```diff
@@ -289,3 +295,19 @@
 
 $subject_prefix_for_site_mails = "OCXX";
 $subject_prefix_for_reviewers_mails = "";
+
+
+// keywords in every header
+$config['header']['keywords'] = 'geocaching, opencaching, skarby,'.
```
deg-pl: :+1: Very good idea!

deg-pl:
1) OC PL and OC UK ready (have $config['ocNode'])
2) What is preffered $config['ocNode'] setting on devel site?
3) Does this PR fix #958 and #1137?
4) Can we remove "oc_on_all_pages_top_1A" and "oc_subtitle_on_all_pages_1A" from translation files?
5) Can we remove nonlusers.txt after merge?

deg-pl:
6. Could you also change meta name="keywords" to new config value - in main.tpl.php mini.tpl.php

In mini.tpl.php we have not HTML5 but XHTML 1.0 Transitional declaration :(

Utils/View/View.php
kojoty: ech - widzisz - niereformowalny ze mnie gość...

kojoty:
> 1. OC PL and OC UK ready (have $config['ocNode'])

cool, thanks!

> 2. What is preffered $config['ocNode'] setting on devel site?

it's depends of what node you want to emulate...

> 3. Does this PR fix #958 and #1137?

sure :) - good to know - I've forgotten about these issues.

> 4. Can we remove "oc_on_all_pages_top_1A" and "oc_subtitle_on_all_pages_1A" from translation files?

I'm going to refactor translations in main layout at the end - it will be done. 

> 5. Can we remove nonlusers.txt after merge?

Yep. There is a change: there was two files in `<dyn-dir>/html/` now this is one json file in `<dyn-dir>.` I thinking if this is neccessary at all - this is a case of one sql query - I'm not sure if we need this at all..

> 6. Could you also change meta name="keywords" to new config value - in main.tpl.php mini.tpl.php

It will be done.

andrixnet:
In the context of https://github.com/opencaching/opencaching-pl/issues/958 I believe `$config['ocNode']` is a poor choice. 
There already is `$oc_nodeid` which is numeric, which will be moved under $config (see my configuration refactor), with a different meaning and scope. 

`$config['ocNode']` does not mean **ocNode** but _node language_.

I had this in mind for the settings refactor: 
```
// Default language for the website. (2 letter language code)
$config['node']['lang']['language'] = 'pl'; 
// Default country. (2 letter country codee)
$config['node']['lang']['country'] = 'PL'; 
// Default locale setting
$config['node']['lang']['locale'] = 'pl-PL'; 
// Date format (PHP definition)
// FIXME ??? shouldn't this be taken from locale?
$config['node']['lang']['date_format'] = 'Y-m-d';
// Time format (PHP definition)
// FIXME ??? shouldn't this be taken from locale?
$config['node']['lang']['time_format'] = 'H:i';
// Date and Time format (PHP definition)
// FIXME ??? shouldn't this be taken from locale?
$config['node']['lang']['datetime_format'] = 'Y-m-d H:i'; 
```

with `$config['node']['lang']['language'] = 'pl';` having the same meaning as `$config['ocNode']`.

andrixnet:
Nevertheless, `$config['ocNode'] = "ro";` set at OCRO in settings.inc.php

kojoty:
@andrixnet,
I think this is a "deeper" problem of names for variables :)
I think that we both are partially right :)

You're right that we have $oc_nodeid, and you're right that is looks similar to language code.

But lets look at my example:
I want to allow site-menu configured per node - so I want to make files:
```
/Config/menu.default.php
/Config/menu.pl.php
/Config/menu.ro.php
/Config/menu.nl.php
...and so on
```
In the code I need to include one file - how to decide which one?
I need to get this node-identifier part like "pl" or "ro". Yes, I can create a simple translator: numericNodeId -> humanReadableNodeId but for me this is a wrong way.

Numbers are good in many places - but they have one disadvantage - people doesn't remember the meaning of numbers if they mean something different than just the number.

In my opinion $oc_nodeid encoding as a numbers was a poor decision. It is used in many places now so we will need to stay with it but it is difficult to understand for people (computer code is for people - not for computers!)

If we need numeric-node-id in the code we should translate it form human-readable form!
In config we should have "human-readable forms".

Then my `$config['ocNode']='pl'` means: "this is OCPL node". Yes, you right - it is looks like language symbol and can be a source of mistakes. Maybe it would be better to use something like for example
`$config['ocNode']="OCPL"` - but it is in my opinion too complicated. 

I believe `$config['ocNode']='pl'` is a good compromise. In future I think we should remove `$oc_nodeid` from config and just translate `$config['ocNode']='pl'` => 3 etc., and add another config switch for devel setups - but this is different discussion...

kojoty:
@andrixnet, I better check the purpose of #958 and you're completely right - I didn't understand it when I wrote comment above - this is not a good fix for #958 - it should be fixed by `$config['node']['lang']['language']` - I will change it assume to your suggestions.

PR #1334, opened 2018-01-14
Fast fixes

kojoty:
This PR contains two fast fixes:
- **DB exceptions in badge_map.php** - occurs from time to time - @tripper1971: please review - i'm not sure if this is a deeper error or it works well - I've just skipped queries which can't run without data

- log warning from **tm_ll_lib.php** - I've often seen it in error.log - This is complicated code which convert coordinates from lat-lon system to  [UTM](https://en.wikipedia.org/wiki/Universal_Transverse_Mercator_coordinate_system). I'm not sure if we even need coordinates in UTM (Has anyone heard of using it in geocaching?) 
It was presented in "coordinates popup":
image

I decided to remove this popup at all and add JS to switch lat-lon coords format on click on coords at viewpage.

Any comments to this?


kojoty:
@harrieklomp, @andrixnet: please not - this PR contains one small translation change which you should probably notice - existing phrase change its meaning a little... - please take a look.

harrieklomp:
Dutch and German translations are correct like it is.

andrixnet:
RO string is fine.

PR #1357, opened 2018-01-21
Fast fixes for startPage + refactoring around geocoding

kojoty:
@andrixnet, @harrieklomp, @deg-pl,
Please note:
Right now oc code has 3 solutions for reverse-geocoding ( region from coords ) - all 3 are presented only at region popup right now - but in further refactoring of cache edit I'm going to use it all 
- NUTS - local data in DB - but work only for EU
- Google Geocode - needs google API key in config and has limited usage because of licence!
- MapQuest geocode - this is NEW - to work it needs mapQuest API key in config - see settingsDefault for details (@deg i add key for oc.pl settings - but not for oc.uk)

There are also small translation changes.


kojoty:
@deg-pl: poprawiłem ładowanie feedów - o czym rozmawialiśmy i teraz są generalnie bez ajaxa o ile są w keszu

Issue #1369, opened 2018-01-23
write_totalstats missing? 


andrixnet:
I get many errors from the cron that `write_totalstats.inc.php` is missing. 

Perhaps something has changed recently and I missed that info, but now searching through GitHub I only got one result: https://github.com/opencaching/opencaching-pl/blob/73c5e0af6bb1620b5ea9d4d01bcec8f1d1ee42f1/docs/sql/crontabs/cron-job-list

Furthermore, from the 5 scripts listed in /tpl/stdstyle/etc in the above, now there seem to be only 2.

Please clarify.
If necessary, please document such changes.

kojoty:
@andrixnet write_totalstats.inc.php is no more use and should be deleted from source. I will update cron doc today evening.

kojoty:
cron docs updated. BTW. I think there will be no scripts in /tpl/etc - I think all such stuff should be stored in mem-cache and generate from code on request.

PR #1372, opened 2018-01-23
#1369: updates in cron docs

kojoty:


deg-pl:
OC PL and OC UK cronjobs table updated

@harrieklomp @andrixnet ping ;)

PR #1392, opened 2018-02-05
Fast fixes

kojoty:


lib/languages/en.php
```diff
@@ -2748,6 +2748,7 @@
     'admCs_cacheSet'  => 'GeoPaths',
     'admCs_currentRatio' => 'Active caches ratio',
     'admCs_requiredRatio' => 'Required caches ratio',
+    'admCs_emptyList' => 'WOW! There is no active but degraded geopaths to list here!',
```
deg-pl: is -> are (geopathS)

kojoty: fix delivered with #1395

PR #1395, opened 2018-02-06
composer + many minor fixes

kojoty:
- en translation fast fix
- kint update by composer
- htmlpurifier updated and moved to composer
- and more...


kojoty:
@andrixnet, @deg-pl, @harrieklomp 

in this PR I updated kint lib and added htmlpurifier to composer

What do you think about removing this code from repo?
I guess you know how composer works, but if not:
 - this is just something like packet manager for php apps - there is simple config (in composer.json file - which we will have in repo).
- to download all necessary external code you need just call `composer update` on local server

I don't know if this is better solution but it means that our repo can contains much less files of external apps. 

What do you think.
(I asked because I'm going to migrate more code of external apps to composer)

deg-pl:
I don't know.
/vendor directory should not be in repo
BUT
Now we have different versions of PHP, DB, Apache, different configs (https, http/2 etc). This makes difficult to create new, modern code and debug problems. When /vendor will not be in repo, every site may have different versions of composer components.
So... I don't know which solution is the best solution.

kojoty:
OK, removing code from /vendor is an easy step to do in future - lets stay with the code in repo

andrixnet:
DB structure and it's predefined contents is hard enough to handle. 
This however would mean adding software components on which the code depends as **an additional layer to maintain for each node**, at each update from the REPO. 
Having the nodes work over a variety of platforms and versions (neither too wide, nor too strict) is a good thing, while a too strict requirement would be in fact unjustified given the scope of OC as an open-source project. 

As developer, I believe functionality such as `composer` is very good in helping maintain all the dependencies properly.

As sysadmin I believe the entire website [code] (including PHP, JS and such dependencies) should be treated as a whole. Ensuring system software and library dependencies (ie PHP version, GC library, etc) are the tasks for the sysadmin.

_Imagine updating a WordPress blog via it's native update function, but then having to run additional library/plugin updates using a different tool, but only for that blog, because these are dependencies but still code updates._

---------------

My recommendations are: 
- specify underlying system software requirements as needed
- provide the website code as a whole and complete package.

PR #1398, opened 2018-02-07
#1374: note popup fix

kojoty:


tpl/stdstyle/js/lightPopup/lightPopup.js
```diff
@@ -1,31 +1,23 @@
 /**
- * This function displays div with given id as popup.
+swdf * This function displays div with given id as popup.
```
rapotek: what is 'swdf'? I'm just curious.

kojoty: it seems this is message for you from my cat walking on my laptop :)
It was in comment so I don't notice it.

PR #1427, opened 2018-02-26
Error/exception global handler

kojoty:
This PR code should catch php errors an exceptions and display custom error page to user + create notice in error log.
I tested it at php7 and it seems to work OK and I hope it will works as well at php5.


andrixnet:
Will return with feedback

PR #1444, opened 2018-03-06
New My Neighbourhood on OC

deg-pl:


deg-pl:
Waiting for @rapotek with his work on new logtype icons

rapotek:
I feel called up to the blackboard... I am not an artist so the icons will not be new by design, only by format: I try to imitate png images in svg as close as possible and see if it looks tolerable if scaled up to large. 9 of 17 icons are ready, more or less. The rest, including the hardest one, still in progress.

rapotek:
@deg-pl here is [the first set of svg log images](https://github.com/rapotek/opencaching-pl/commit/e1c64abdd4df5c88935cc8085b2a1b30ab0a382d), two are still missing: note-old and Arrow-Right, but I think these two are not necessary for new MyNbh. They still needs a work I suspect, but check them out, if they are correct, you can add them to this PR.

deg-pl:
SQLAlter added in this PR is my mistake. Please don't run it. Today I'll remove it from repo.

andrixnet:
@deg-pl here it is: 
```
[Sat Mar 10 14:11:35.368791 2018] [:error] [pid 31655] [client 193.231.72.224:54389] PHP Fatal error:  Cannot use isset() on the result of an expression (you can use "null !== expression" instead) in /serve/www/oc_ro/oc_www/lib/Objects/User/UserPreferences/UserPreferences.php on line 124
```

deg-pl:
@andrixnet Thanks! This was an older error. I fixed it and now new MyNeighbourhood works also on OC RO. From OC UK FB group:

The new My Neighborhood
Assumptions that guided its design:
* It is possible to define more than one area
* Banally simple configuration of the Neighbourhood - departure from entering cords and radius for drawing on the map
* Large, clickable fields for easier operation on touch screens
* The greatest possible personalization
* A more developed help system and hints
Has the assumptions been completed? Evaluate yourself. Certainly you will need to change a few habits, but I hope that it is worth it. If anyone has a problem with it - I have prepared a "minimalist" mode.

What will change in the future? Two things:
* Visualization on the map of the cache / log which you will point with the mouse. For now, everything is only thrown on the map, but with a large amount of cache, it is less known which is which. Here I am waiting for the code to be written by Kojoty (OC PL).
* The possibility of receiving notifications from not only my main Neighbourhood area, but also additional ones. A small revolution in notifications soon ...;)

PR #1446, opened 2018-03-07
DB cleanings

deg-pl:


harrieklomp:
Run "sql 2018-03-06_add_notify_to_additionalnbh.sql" on OCNL. Will do other when we are on new server.

andrixnet:
Both on OCRO. Thank you.

PR #1461, opened 2018-03-19
User management refactoring

deg-pl:


deg-pl:
Was (2-3 weeks ago):
image

Will be:
image

deg-pl:
@harrieklomp @andrixnet There are some translations - keys admin_user_* and admin_notes_* - all of them are marked by `/* EM */` comment.
They are not so important because these texts are for admins only.

PR #1476, opened 2018-03-31
Add French translation

kumy:
This is my first pass on translation to French using, probably, outdated reference file.


deg-pl:
WOW!!! Thanks!
Rememer that FR lang file has many strings missing. Compare it with EN lang file.

kumy:
Translation in French completed. You can merge it if you wish.

kojoty:
@kumy, Thanks for that.

Issue #1464, opened 2018-03-23
Meeting the requirements General Data Protection Regulation Law


deg-pl:
From the 25'th of May 2018 we have to meet requirements of the General Data Protection Regulation Law (https://www.eugdpr.org/). We have specific guidelines prepared what needs to be changed in the code and we will change it. **But I'm nearly sure, that ALL of OC nodes should change their Terms of Use**. From 25.05.2018 our code will enforce acceptance of new Rules (or will deactivate account).

@harrieklomp @andrixnet @Amberel Do you have new rules? Are work on it ongoing?

harrieklomp:
I have some questions and perhaps some suggestions.

- Will there be a standard Terms of Use for all nodes? If yes, this could be in English and must be translatable.
- Are there separate Terms of use for the main site and for the forum? Most nodes do have a forum and/or blog etc.
- Are other nodes be informed about the progress? Perhaps visualisation on Deg's development site?

deg-pl:
@harrieklomp 
Our OC Team is working on changes in the regulations. I am afraid that each country will need separate regulations. GDPR is Europe-wide law, but every european country has executive law to meet local standards. For example - GDPR defines the range between 13 and 16 years of age when parents' consent for the use of service is required. The Polish law defines this age as 13. I do not know how it is in your's country. OC PL regulations contain many references both to Polish law and the specificity of OC PL (i.e. we has no virtual caches, our OC Team consists of a Technical Team and a 16-person Centurm for Geocaher Service)

BUT :) our OC Team prepares document with most important changes - in English, and will share it with other OC nodes.

andrixnet:
We will have to research this in depth.

harrieklomp:
On OCNL i have made a new wiki page with GDPR guide (in Dutch) on http://wiki.opencaching.nl/index.php/Privacyverklaringen. This is done because there is a lot to explain to user about this subject and it is not recommended to use some lines in only "Terms of Use". There are a lot of standard forms on internet who does explain (Polish, English or Romanian) witch have all text in it to use for this. Perhaps only change sitenames  on this form.
I would like to suggest to use on registration page two links "Terms of Use" and "Privacy statement". This "Privacy statement" should also added to footer.

harrieklomp:
[...]

PR #1540, opened 2018-05-20
Next changes related to GDPR (#1464)

deg-pl:
@andrixnet @harrieklomp Please - review and translate strings beginning "register_". There is also an SQLAlter to reset previeous confirmation of rules. Please do this SQLAlter BEFORE 25 of May.

@andrixnet @harrieklomp @Amberel - have you new GDPR rules (terms) ready? From 25 of May all users will have to reconfirm read new rules (terms), or they can't use OC, and may disable their accounts via OC Team.

### DIFF
--- a/Controllers/UserRegistrationController.php
+++ b/Controllers/UserRegistrationController.php
@@ -104,6 +104,7 @@ class UserRegistrationController extends BaseController
         $this->view->setVar('username', $username);
         $this->view->setVar('email', $email);
         $this->view->setVar('errorMsg', $errorMsg);
+        $this->view->setVar('minAge', $this->ocConfig->getMinumumAge());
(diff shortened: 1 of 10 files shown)


deg-pl:
@andrixnet @harrieklomp @Amberel - GDPR assumptions minimum age minimum age to register without parents' agreement to 16 years old. BUT goverments of EU countries can set their own limit. In Poland we have 13 years old limit.
Default OC configuration is 16 years old. If you want to change this limit, please add to settings.inc.php something like:
`$config['limits']['minimum_age'] = 13;`

harrieklomp:
Translated "register_" string to Dutch and run sqlalter on OCNL.
New GDPR rules on OCNL wiki is almost done.

PR #1557, opened 2018-06-07
Date published unknown if before date_created

rapotek:
This is a fast fix for date of publication problems for historical caches. On OCPL despite an update algorithm it seems the date of publication is set to the date of sql alter update time instead of the computed times for existing caches.
If the situation occurs the following sql should be applied:
`UPDATE caches SET date_published = DATE_SUB(date_created, INTERVAL 1 DAY) WHERE date_published IS NOT NULL AND date_published <= '?'`
where `?` should be replaced by the date and time of `date_published` introduction (possibly '`SELECT MIN(date_published) FROM caches`' result).

BUT REMEMBER: FIRST CHECK IF THE SITUATION DESCRIBED ABOVE OCCURS!


rapotek:
I have no time for now to monitor this. In any problems or questions I will be available in the evening,

rapotek:
It seems the date of publication has been corrected on OCPL, so is this PR still vaild then, or should it be closed?

PR #1575, opened 2018-07-07
Mapv4 - next changes

kojoty:
this is small next step in mapv4 preparation


kojoty:
I merge in weekend but this is side-work for the service and I need it for further works

Issue #1521, opened 2018-05-06
Multiple opencaching-pl instances cannot be hosted on the same server


andrixnet:
Apparently two or more instances of **opencaching-pl** code cannot run simultaneously on the same server because some caching machanisms overlap.

At least "newest caches" section on homepage is cached as a code fragment under filesystem /tmp, rather then oc-dynamic/tmp.
This results in one instance displaying cached data generated by another instance.

I have created two instances, one for OCRO, one for OCNL, as tests here: 
http://ocro.dev.andrix.eu/
http://ocnl.dev.andrix.eu/

Another one for OCUS will follow. 

I believe it can be a real scenario that two OC instances be hosted on a single server at a point in the future.

andrixnet:
There is a `tmp/` directory defined under oc-dynamic in settins.inc.php. It could/should be used for such operations.

deg-pl:
Reopen. OcMemCache is incompatibile with multiple OC instances on simple machine.
const OcMemCache::KEY_PREFIX should be configurable / related with site name (@kojoty)

kojoty:
@deg-pl myślę, że można dodać po prostu konfigurowalny prefix (z configs) dokładany do nazwy klucza, pusty jako default...

deg-pl:
Ok. Done (after merge).

PR #1599, opened 2018-08-22
Migration of all locks to BaseDynDir (#1521 related)

deg-pl:


rapotek:
Checked it locally, seems to work: locks, notifications data and feeds are saved under dynamic files directory.

PR #1613, opened 2018-09-05
Failover language added for tr and tr2 functions (fixes #1594)

### LINKED ISSUE #1594: When no translation, use English translation
Because programmers always use English and Polish, so why not use automatic English translation when there is no other translation available?

When translating hardcoded (Polish) parts we always need to translate and add strings in all available language files. This problem is also when adding new code with translations.
When there is a string missing in a language the line "No translation available (identifier: this_is-a_translation-line)-todo" shows up.

This means less work for programmers when translating and less mistakes between different language files because programmers only use these 2 languages. Also programmers do not need to wait for other translators.
When new strings are added into "en.php" i upload it to Crowdin (can be done automatic) and every translator can see the new strings without scrolling all the way down in the specific language file.

rapotek:
Adds the solution of failover english language to translations, that means if translation key is missing in currently selected language, the 'en' translations array is loaded if needed, checked for the key and if the key is not found there too, "No translation available(...)" is returned. Therefore adding new keys to every living translation file is no more necessary, providing the new keys are added to 'en' translation file. The missing keys can be added some time later.

The translations initialization has been moved from `common.inc.php` to `language.inc.php`, too.

### DIFF
--- a/lib/common.inc.php
+++ b/lib/common.inc.php
@@ -3,15 +3,10 @@
 require_once __DIR__ . '/ClassPathDictionary.php'; // class autoloader
 
 use Utils\View\View;
-use Utils\Uri\Uri;
-use Utils\I18n\I18n;
-use Utils\I18n\Languages;
 use lib\Objects\ApplicationContainer;
 use lib\Objects\User\User;
 use lib\Objects\User\UserAuthorization;
 use lib\Objects\OcConfig\OcConfig;
-use Utils\Text\UserInputFilter;
-use Utils\Uri\OcCookie;
 
 session_start();
 
@@ -68,7 +63,7 @@ if (php_sapi_name() != "cli") { // this is not neccesarry for command-line scrip
     UserAuthorization::verify();
 
     initTemplateSystem();
-    loadTranslation();
+    initTranslations();
 
 }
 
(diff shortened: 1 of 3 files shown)


lib/common.inc.php
```diff
@@ -125,6 +125,15 @@ function loadTranslation(){
             $lang = OcCookie::getOrDefault('lang', $lang);
         }
 
+        // load failover language settings
+        if (
+            !empty($config['failoverLanguage'])
+            && !isset($language[$config['failoverLanguage']])
+            && I18n::isTranslationSupported($config['failoverLanguage'])
+        ) {
+            load_language_file($config['failoverLanguage']);
```
kojoty: hmm... I have a thought that maybe load "fail over" language only if there is a lack of translation...
Generally - in most cases - we should have all the needed translations so there is no need to load these data for almost every script in OC.

kojoty: BTW - I think that we need to move this code to language.inc.php file - I would like to make common.inc.php as small as possible - maybe just add some "translation_init()" mehtod to common.inc.php and move the rest to the language.inc.php.

lib/settingsDefault.inc.php
```diff
@@ -73,6 +73,7 @@
     'supportedLanguages' => array (
         'pl', 'en', 'nl', 'ro'
     ),
+    'failoverLanguage' => 'en',
```
kojoty: I'm not sure if failover lang should be configured at all - if we have only one failover let's assume that this is english, lets hardcode it as const in language.inc.php - I can't imagine who of us needs to change it to any other lang...

lib/common.inc.php
rapotek: You are right, I didn't think it could be done this way. I will try to rewrite it and update the PR description.

lib/settingsDefault.inc.php
rapotek: Well, I hold with hardcode as less as possible, optimally nothing at all. Of course, there can be a `FAILOVER_LANGUAGE` constant set to 'en' but won't it be a step back?

kojoty:
@rapotek, a powiedz jeszcze - Ty pewnie testowałeś tę zmianę z różnymi językami = rozumiem, że sprawdziłeś, że to działa OK - można to mergować w ciemno...?

rapotek:
@kojoty sprawdziłem to tak: dodałem testowo jeden klucz do polskiej i angielskiej wersji, widoczny w stopce przed flagami, w pozostałych (nl i ro) tego klucza nie było. Efekt: dla wybranego polskiego języka pojawiała się polska wersja, dla pozostałych - angielska. Nie było nigdzie 'No translation available(...)'.

Tu masz przykłady:
failover_language_1
failover_language_2
failover_language_3
failover_language_4

PR #1617, opened 2018-09-11
fixed OKAPI search for found caches

following5:
The last automatic OKAPI update in May worked, but this one didn't, so again a manual pull request.

@kojoty 
You may contact me at following@online.de so that we can finish the OKAPI update setup via your server.

### DIFF
--- a/okapi/services/caches/search/SearchAssistant.php
+++ b/okapi/services/caches/search/SearchAssistant.php
@@ -170,20 +170,23 @@ class SearchAssistant
                 $tmp = substr($tmp, 1);
                 $operator = "not in";
             }
-            $types = array();
+            $sizes  = array();
             foreach (explode("|", $tmp) as $name)
             {
                 try
                 {
                     $id = Okapi::cache_size2_to_sizeid($name);
-                    $types[] = $id;
+                    $sizes[] = $id;
                 }
                 catch (Exception $e)
                 {
                     throw new InvalidParam('size2', "'$name' is not a valid cache size.");
                 }
             }
-            $where_conds[] = "caches.size $operator ('".implode("','", array_map('\okapi\core\Db::escape_string', $types))."')";
+            if (count($sizes) > 0)
+                $where_conds[] = "caches.size $operator ('".implode("','", array_map('\okapi\core\Db::escape_string', $sizes))."')";
+            else if ($operator == "in")
+                $where_conds[] = "false";
         }
 
         #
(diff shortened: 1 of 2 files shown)


kojoty:
@following5 - we definitely needs to fix the automatic okapi update - I'll try to take a look at this today evening - direct commits to OKAPI is not a good idea in my opinion

following5:
There is effectively no difference between distributing OKAPI code via the automatic script or this way via pull request. I commit to the OKAPI repo, and then do the replication to OCPL code manually. This is how OKAPI was distributed to the OCDE for many years, and there never was any problem with that.

It's just more convenient to have it done automatically.

kojoty:
> It's just more convenient to have it done automatically.

sure, this is also safer because it is not dependent of whether you remember or not to commit to OCPL repo - Let me check if automatic update is easy to restore 

BTW. @following5 do you know who else (besides OCPL) has using the OKAPI tile service to display map of caches?

following5:
I am not aware of anyone besides of OCPL to use the OKAPI map. But it may be an option for OCDE to fix it's broken map.

harrieklomp:
The last working automatic update was on 29th of may. Please look at https://github.com/opencaching/opencaching-pl/commit/3b4714632687f0cb72efe8346f513e142861ca4b
Perhaps looking from this date to start investigate to find the problem.

kojoty:
Thanks Harrie, I more-less know why this happen - there is a problem at private Wrygiel's server, which was a host for okapi versioning and was call by okapi GitHub hooks. I need to migrate that service somewhere - the problem is there is only the script in Python - so I need to find web server with Python support or rewrite it to php. 

@following, @harrie - maybe you have some server with Python 2,7 where I can migrate that service?

This is quite sure simple - that Python script needs to generate the archive with Okapi, increase the version and call the  OC* servers...

following5:
Umm, yes, I have a server with python 2.7. Will give it a try.

following5:
I am sorry, but I failed in making it run on my server. There is a bunch of JSON data in the python script with some data of @wrygiel; maybe this needs to be replaced.

The script on Wrygiels server is inactive now due to changed webhook settings. You may test your webhook setup in your private fork of the OKAPI repo (after updating that to the current OKAPI code), so we don't need to interact for testing. When your setup works, we can activate your script in the OKAPI repo.

Last line of OkapiUrls.php is used for commit tests.

kojoty:
@following5, May I ask you to prepare PR to OCPL code if there is any important fix in OKAPI code. In the meantime I will prepare okapi autoupdater based on wrygiel's script. I analyzed it yesterday and it seems it should be quite easy to port - but I will do it when we finish the works on switch from Google Maps to Openlayers API what makes me busy last days.

One more question - how OCDE deals with OKAPI? - I see there is no OKAPI source in the OCDE source code.

Thanks!

following5:
Okay.

OCDE deploys OKAPI via PHP Composer. The updates are manually triggerd by running a script which updates https://github.com/OpencachingDeutschland/oc-server3/blob/development/htdocs/okapi/meta.php. OKAPI code then will be checked out directly from the OKAPI repo to the Opencaching.de server.

following5:
I don't know how it works in detail; this is maintained by someone else, and I am no longer member of OCDE development team.

kojoty:
OK, I found it in OCDE code: https://github.com/OpencachingDeutschland/oc-server3/blob/development/dev-ops/local.team-opencaching.de/OkapiHelper.php

PR #1632, opened 2018-09-15
OKAPI r1708

following5:
This fixes the OKAPI error notifications

- "A non-numeric value encountered" in replicate/changelog
- "SQL Error 1241: Operand should contain 1 column(s)" in search/bbox

and some more bugs.

### DIFF
--- a/okapi/OkapiUrls.php
+++ b/okapi/OkapiUrls.php
@@ -37,4 +37,4 @@ class OkapiUrls
     );
 }
 
-# This line is used for commit-hooks testing: ............>
+# This line is used for commit-hooks testing: .............>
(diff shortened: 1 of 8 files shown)


kojoty:
Thanks @following5!

following5:
Code looks fine, and I could not reproduce the problem. The errors occured only in the same minute when you merged this PR.

Looks like a temporary problem during code deployment. Is the Opencaching.PL webserver shut down during deployment? If not, that would explain the problem.

kojoty:
I think so - we usually don't stop services for update so it is possible that somethings fail during the update - but it shouldn't be danger (usually).

deg-pl:
@following5 we have problem with OKAPI. When we get caches via c:geo or "Awaryjniejszy GPX", OKAPI gives caches not near selected coords, but random caches - spread acros the country.

following5:
I am sorry for that. Should have tested more carefully.

Issue #824, opened 2016-11-07
Volunteer requirements = configurable


andrixnet:
Make volunteer cacher criteria as configuration variables instead of hardcoded parameters.

kojoty:
@andrixnet - is that what you wanted to get?
See /Config/guides file.

Please reopen if there is anythings more to do.

harrieklomp:
Sorry that i reopen this issue.

Should translation files not show the the "set value"?
Now it is hard coded "20" in all translations.

kojoty:
I hope now it can be marked as done

PR #1658, opened 2018-09-19
#824: configurable guides requirements

kojoty:
@andrixnet, @harrieklomp please notice the language files change in this PR

+ 
@deg-pl, @rapotek : since now it is possible to pass arguments to tr() function like to vsprintf() function 
- see this PR for details.


deg-pl:
Great! It solves some problems with translation files.

PR #1659, opened 2018-09-20
OKAPI r1712

### LINKED ISSUE #1502: Attend event in future
Perhaps it could be possible not to show logtype(s) "Attended" and/or "Maintenance performed" for a event before event has taken place.
Only "Comment" and "Will attend" should be visible.

selectie_167

following5:
This fixes #1502/#1649 in OKAPI for 'Attended' logs. You may want to add PL/NL/RO translations for the new message.

I considered to verify also 'Will attend' and 'Found' dates, but that's tricky. If an event takes multiple days, 'Will attend' may be logged after the listed event date. And at least at OC.de there are caches with wrong hide dates and inactive owners (e.g. OC6877), which could  no longer be logged. Probably there are many GC listings which were copied later to OC, but without properly backdating the "hidden" date. Blocking "Found it" before hide date would block users from transfering their old GC logs to those OC caches.

But if you still want such checks for OCPL, I could add them for OCPL only.


kojoty:
@following5 
Am I right that there is only one new sentence in translations:
```
"You cannot attend an event before it takes place. Please check the log type "
"and date."
```
If yes the polish phrase would be:
```
"Nie możesz zalogować uczestnictwa w spotkaniu, zanim to się odbędzie. Sprawdź typ logu"
"i datę."
```

kojoty:
@mzylowski what do you think about @following5 considerations?

mzylowski:
`Nie możesz zalogować uczestnictwa w spotkaniu, zanim się ono odbędzie. Sprawdź typ logu i datę.`

I fully understand your considerations, but it is kind of consensus. For my point of view 'will attend' log is important for event owner to know how many people is interested, how many people will be on the event and etc. In the day of event starting, additional logs 'will attend' doesn't change anything. Everything is prepared. 'will attend' log doesn't count to any stats etc. It's only information, like comment.

PR #1675, opened 2018-10-04
OKAPI r1715

following5:
- translation update & cleanup
- fix error "Incorrect datetime value: '0'"


following5:
Looks like some unusual strict type checking has been activated on the OC.pl SQL server, which produces all these errors. Can you disable that?

deg-pl:
This is default MariaDB config. On the old server we had strict checking manually disabled, on the new one we prefer to not disable it (if it is possible).

following5:
Lots of different errors. Would need a major review of OKAPI code to make it stable in strict mode. Unfortunately I can't dedicate enough time to this. I may just try to do some more hotfixes for the errors as they occur.

PR #1683, opened 2018-10-06
OKAPI r1716

following5:
Okapi says "undefined offset: 0" when data of a non-public geocache like OP8UPQ is requested. I am not sure how that happens and cannot reproduce it, but this may fix it.

### DIFF
--- a/okapi/core/Okapi.php
+++ b/okapi/core/Okapi.php
@@ -22,8 +22,8 @@ class Okapi
     public static $server;
 
     /* These two get replaced in automatically deployed packages. */
-    private static $version_number = 1715;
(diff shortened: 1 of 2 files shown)


following5:
Didn't work.

```
            $rs = Db::query("
                select cache_id, language, `desc`, short_desc, hint
                from cache_desc
                where cache_id in ('".implode("','", array_map('\okapi\core\Db::escape_string', array_keys($cacheid2wptcode)))."')
                and cache_id <> 0   /* workaround for issue #540 */
            ");
```
cache_id = 0 is explicitely excluded. But still we get an "undefined offset: 0" here:

```
            while ($row = Db::fetch_assoc($rs))
            {
                $cache_code = $cacheid2wptcode[$row['cache_id']];
```

Issue #1697, opened 2018-10-14
geo-coding and search for the main map :/


kojoty:
Search or the mainMap - at first: 
- looking for geographical location by given name

PR #1698, opened 2018-10-14
#1697 - search for the map - initial version

DorotkaMakotka:
@harrieklomp, @andrixnet - please note that this PR contains some new phrases that need to be translated.

@harrieklomp , @andrixnet, @deg-pl - search needs OpenRouteService api_key - see Config/map.default.php

This is initial version of the search mechanism - for now main map only.

Plans for future iterations:
* handle quota (counting usage, perhaps switching between geocoding providers)
* search by geocache waypoint code/name
* jump to given coordinates
* ?


kojoty:
please don't merge this (and any other) PR - OCPL git repo at production server has temporary manual  fixes in OKAPI code

tomorrow I will clean up it

lib/Objects/Coordinates/GeoCode.php
```diff
@@ -14,6 +18,78 @@ class GeoCode
...
+        $config->getMapConfig();
+        $ors_key = $config->getMapConfig()['keys']['OpenRouteService'];
+
+        if(empty($ors_key)) {
+            Debug::errorLog("No api_key for OpenRouteService in configuration. Check /Config/map.default.php");
+            return null;
+        }
+
+        $input = urlencode($place);
+        $url = "https://api.openrouteservice.org/geocode/search?api_key=$ors_key&text=$input";
```
rapotek: Just an idea, I had no time to check:

1.  In `mainMap.js`, `_getLocalizationByPlace()`: `var center = ol.proj.toLonLat(map.getView().getCenter());`, then pass it to this method
2. Append to the url `focus.point.lat=<lat_from_center>&focus.pointlon=<lon_from_center>`
3. Sort the result features by feature.properties.distance ascending

Then you will have as the first list item the nearest place to the current view center.

PS: remember anyway, the default size of returned features is only 10, it may be too small to return all places with a popular name.

DorotkaMakotka: Current state is very much initial, just to enable searching at all. For sure there will be next version soon, and I will certainly think over your suggestions while preparing the next iteration. Thanks!

deg-pl:
OC PL and OC UK ready!

harrieklomp:
API key added on OC NL

tpl/stdstyle/mainMap/mainMap.js
```diff
@@ -518,5 +519,140 @@ function initControls(params){
...
+    var _applySelectedPlace = function(result) {
+        _adjustMap(result);
+        closeSearch();
+    }
+
+    var _adjustMap = function(result) {
+        var bbox = result.bbox;
+        var sw = ol.proj.fromLonLat([bbox[0], bbox[1]]);
+        var ne = ol.proj.fromLonLat([bbox[2], bbox[3]]);
+        map.getView().fit([sw[0], sw[1], ne[0], ne[1]], {nearest: true});
```
rapotek: I know it is a little bit too late, but I recommend to use:
`map.getView().fit([sw[0], sw[1], ne[0], ne[1]], {nearest: true, maxZoom: map.getView().getZoom()});`
instead.
The zoom should then stay as is was before search result list element click.

PR #1732, opened 2018-10-30
Rootpath refactor

kojoty:
@deg-pl @andrixnet @harrieklomp 

This PR contains a few things:

- you can remove APC-related password from settings.inc.php - now APC is accessible for users with sysAdmin role under **/sys/apc** (see new sysController)

- testController (/test) is now accessible only for users with advUser role

- I made "big" cleanup: I removed almost all $rootpath and $stylepath variables from almost whole code - in my opinion dealing with these global vars was too complicated - I hope now this is better - please review as carefully as you can :) - there are lot of changes and I couldn't test all the scripts...


deg-pl:
@kojoty WOW! You put a lot of work into it. I think we need easy interface to manage user roles.

I think - @andrixnet @harrieklomp @kojoty and me should have "sysAdmin" role on all OC nodes. We all should may do tests new features across nodes or for example - check PHP version (you can do it in APC) or cache functioning.

okapi_settings.php
```diff
@@ -25,7 +25,7 @@ function get_okapi_settings()
     # OKAPI defines only *one* global variable, named 'rootpath'.
     # You may access it to get a proper path to your own settings file.
 
-    require($GLOBALS['rootpath'].'lib/settingsGlue.inc.php');  # (into the *local* scope)
+    require('./lib/settingsGlue.inc.php');  # (into the *local* scope)
```
rapotek: @kojoty Did you check this? It does not work when making OKAPI call in my dev.

kojoty: @rapotek no właśnie! - i dlatego w kolejnym commicie jest to zmienione - ten PR trzeba brać jako całość - commity mogą nie działać jesli je wyciagniesz osobno :(

rapotek: OK, nie zauważyłem, że wybrałem commit a nie files

kojoty: niemniej nie wykluczam, że coś podobnego pominąłem w kodzie... więc patrzcie mi na ręce!

PR #1738, opened 2018-11-03
update last_found when changing event log type

following5:
Bugfix: When changing the type of an event log from or to 'Attended', caches.last_found was not updated.


kojoty:
thanks for the patch!

PR #1750, opened 2018-11-11
Enable Okapi diagnostics for geopath search

following5:
This enables Okapi developers to evaluate the usage of 'powertrail_ids' search parameter. If noone uses it except OCPL code, we can remove it from the public interface. See https://github.com/opencaching/okapi/issues/560 and https://github.com/opencaching/okapi/commit/e6a78302be86d1867d782d4b0da5e4f76e0dec70 for more information.


following5:
Found way to do this without an OCPL code change.

Issue #860, opened 2017-02-09
New cache types (requested by OCNA)


andrixnet:
The is a list of new cache types requested by OCNA in advance of their migration to OCPL-current:

- BIT Caches _(special QR code label, similar to Munzee)_; ID=12
- MP3 Caches _(same as podcache)_; ID=15; must change at import from id=9
- USB Dead Drop Caches _(as cache type, instead of just attribute)_; ID=11
- Guestbook Caches; ID=18; must change at import from id=10
- Letterbox Caches; _(as cache type, instead of just attribute)_; ID=13 
- Benchmark Caches; ID=19

Caches should be added with ID's according to column F as described here: 
https://wiki.opencaching.eu/index.php/Cache_types

New cache types should go into `forbidenCacheTypes` by default for existing nodes.

Code changes needed (at least): 
- [ ]  cache type definitions (in code)
- [ ] cache type definitions in database (still required)
- [ ] translations
- [ ] default graphical resources
- [ ] default config

andrixnet:
I also suggest using this opportunity to redefine current porentially overlapping IDs in the code and database and have the code define the complete set of cache types, even if only a subset is available by default.

deg-pl:
I think it is not a very good idea. It is source of GIGANTIC problems with incompatibility with 3'rd party software (c:geo and MANY others).

I think better way is migrate these cache types to right cache type and use attributes. For example - USB Dead Drop cache can be traditional cache, multi-cache (http://opencaching.pl/viewcache.php?cacheid=53498), moving cache (http://opencaching.pl/viewcache.php?cacheid=17058), mystery cache (http://opencaching.pl/viewcache.php?cacheid=18287), other type cache (http://opencaching.pl/viewcache.php?cacheid=23419). Dead Drop is type of logbook (attribute!) not type of cache.
Letterbox means that is a stamp in the cache. It can be every cache type!

If I go for caches - I usualy don't search multicaches because them requre much more time. I can't determine easily that Dead Drop or Letterbox is multicache or traditional cache.  In PL we have attribute "letterbox" or "dead drop" - so cache type is adequate to phisical type. Thats a big difference for me.

Opencachingna:
@deg-pl  Most of these cache types were added by our founder in 2010 (he is no longer involved with the site). I think our userbase would be very upset if any of them went away! Generally not too much of a problem with c:geo, they all show up on the maps or in a list, they generally just have a mystery cache icon, rather than their own unique icon.

[...]

PR #1752, opened 2018-11-11
reserve ID 14 for geopath finals

following5:
Someone reserved ID 9 for the (to do) geopath finals. It was marked with "TODO" in GeoCacheCommons.php.

9 collides with OC.US podcast caches and would complicate OKAPI implementation. So I suggest do reserve ID 14 for that, which is the next free global ID.


following5:
After consulting #860, I guess it's too early to decides this. OCUS special types may be dropped, and then more numbers are free, also in OKAPI.

Sorry for those premature PRs. I will shut up now. :-|

kojoty:
BTW I think that OCPL will drop something like "geopath final" soon. Geopaths will not have something like that.

PR #1762, opened 2018-11-16
OKAPI map fix - do not prentend to be another user

following5:
This will fix https://github.com/opencaching/okapi/issues/496. It prevents accidential publication of private data in future OKAPI versions, e.g. showing other user's unpublished caches on the map.


Controllers/MainMapAjaxController.php
```diff
@@ -269,17 +277,27 @@ private function getSearchDataParam()
     /**
      * Parse map filter params and convert it to okapi search params
      */
-    private function parseUrlSearchParams()
+    private function parseUrlSearchParams($userId)
     {
+        Facade::reenable_error_handling();
+
+        $userUuid = \okapi\core\Db::select_value("
```
kojoty: @following5 I understand this means one more sql query for every tail - do we really need it - we can just add user-uuid to JS-ajax request instead of internal userId... am I right?

following5: Compared to the other SQL queries that OKAPI does for each tile - some of them by magnitudes more expensive -, it's neglectable. Will increase runtime by maybe 0,01% or so. ;-)

But if the UUID is already available somewhere else - sure, it may be passed to /MainMapAjax/getPopupData/ and to /MainMapAjax/getMapTile/, so we can elminiate that query.

But, I missed a bug, so please don't merge yet! Will fix.

kojoty: User UUID can be accessed - this is a little bit complicated but:

- data of user for which the map is displayed is passed to JS here:
https://github.com/opencaching/opencaching-pl/blob/master/Controllers/MainMapController.php#L118

- now to MainMap JS is passed only userId and username - but we can pass UUID as well - we need add field to MainMapJsParams

- then in JS instead of userId we can just pass UUID in place where `getMapTile` is called

If you want we can just merge it as is (in Monday) after your next update with mentioned bug-fix.

following5: Okay, I have fixed the bug and eliminated the SQL query.

(The latest commit also includes 3900ca73, so that this branch can be tested without a rebase or merge.)

kojoty:
@following5, 
I understand this is ready to merge (after resolving conflict) - am I right?

following5:
Right! The merge conflict is resolved now.

(And sorry for the error-spam earler today. That was a rare combination of a flaw in my testing workflow and a bug in Okapi anti-spam-mode.)

PR #1769, opened 2018-11-23
Add OKAPI signalling cronjob

following5:
This is a proposal to fix https://github.com/opencaching/okapi/issues/552 - update merit badges when users submit OKAPI logs. Also, it prepares editing cache coordinates via OKAPI, by doing the altitude calculation. And allows to add future "OKAPI->OCPL housekeeping" jobs.

Advantages:
* No redundant logic in OKAPI for merit badges etc.
* No code dependencies - if OCPL developers want to change the MeritBadgeController interface, this can be done independently from OKAPI code.
* Allows to change cache coordinates without having to wait for the external altitude provider. (This feature could also be provided via Facade, if needed.)

Disadvantages:
* The changes are not updated immediately, but with a time lag of some minutes by a cronjob.

The OKAPI side is already implemented and active. At OCPL sites, the following cronjob would be added:

`*/5 * * * * php /var/www/html/ocpl/util.sec/okapi/okapi_cronjob.php`
(or wherever the file is located.)

OKAPI stores the update reqests e.g. for merit badges in the `okapi_signals` table, and the new OCPL `OkapiController` cronjob reads and executes them.

### DIFF
--- a/Controllers/Cron/OkapiController.php
+++ b/Controllers/Cron/OkapiController.php
@@ -1,0 +1,62 @@
+<?php
+namespace Controllers\Cron;
+
+use Controllers\BaseController;
+use lib\Controllers\MeritBadgeController;
+use lib\Objects\GeoCache\GeoCache;
+use okapi\Facade;
+
+class OkapiController extends BaseController
+{
+    public function __construct()
+    {
+        parent::__construct();
+    }
+
+    public function isCallableFromRouter($actionName)
+    {
+        // This controller is used by cron only - router shouldn't call it!
+        return false;
+    }
+
+    public function index()
+    {
+        // The OKAPI signal processing is reentrant - we don't need a lock.
+        $this->process_signals();
+    }
+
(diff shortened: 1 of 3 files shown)


deg-pl:
@following5 We have SimpleRouter - your code can be reached by /Cron.Okapi/ - without calling  util.sec/okapi/okapi_cronjob.php

kojoty:
@deg-pl do you have a time to look at this (if not just give me know please)...

deg-pl:
@following5 sorry for my comment 2 weeks ago. You did it correctly.

Please update /docs/crontabs/cron-job-list yet.

following5:
done

deg-pl:
@andrixnet @harrieklomp There is a change in crontab - one additional job. Even if you don't use merit badges, you should add this to cron because of engine to update caches altitiude after changes in OKAPI.

harrieklomp:
Added cronjob on OC NL

PR #1771, opened 2018-11-25
use configured default languages for GPX

following5:
Sets `caches.default_desclang` from `config['defaultLanguageList']` instead of PL, EN. This field controls the language of GPX download.


following5:
Other solution didn't work.

PR #1774, opened 2018-12-05
fixed username regex and escaping; fixes #1173

following5:
There were some duplicates and some missing escape in the REGEX_USERNAME.


following5:
Main problem was the sequence ` -=` (space minus equals). Within [], it allows any characters from a space until =. So the characters ! " $ % ' , : ; < were acidentally allowed in user names.

kojoty:
@deg-pl , could you take a look at this and merge if this change is OK?
I don't have a time to test it in details - especially we need to resolve the issues around the usernames listed above.

Issue #1254, opened 2017-10-06
Introduce a cronjob engine


following5:
It looks like there are only standalone cronjobs in the OCPL code. If a new cronjob is needed, a new script must be installed in the crontabs of all nodes.

What do you think about adding a cronjob engine like that of OCDE and Okapi? Only one entry in crontab, and new cronjobs are added by creating new class files. For each cronjob, the execution time<sup>x</sup> or interval is specified in the class, and the next execution time of all the jobs is maintained in a database table.

<sup>x</sup> Neither OCDE nor Okapi have fixed execution times, but just intervals. But from [this comment](https://github.com/opencaching/opencaching-pl/pull/1246#issuecomment-333998311) I know that you prefer fixed times in the night for some tasks.

deg-pl:
This is a VERY good idea. We have problem with sync cronjobs across all nodes.

We use some kinds of cronjobs:
- running at fixed time once a day (maintenance at night), or even once per week/month (i.e. "geopath/cache of the week/month").
- running many times a day (with given interval), but we'd like to manage time when job is done. For example - ad ?:00 we publish new geocaches, several minutes later we send e-mails to users with info abount new caches
- running at specified interval - for example GeoKrety management

following5:
The new cron engine is enabled at PL, NL and UK sites.

@andrixnet 
Could you please update the crontab, as shown [here](https://github.com/opencaching/opencaching-pl/commit/69ad770f162487ed65b718f13908d620af0c1390#diff-7a541949c6bd7d0c20559572ce922bda)? This should be the last manual crontab update; future cronjob maintenance will be done via Git.

following5:
Thanks!

PR #1777, opened 2018-12-09
New Cronjob Engine

following5:
This implements a universal cronjob engine, as proposed in #1254. Implements all timing variants as requested by @deg-pl in #1254. The engine is located in `Controllers/Cron/CronJobController.php`, plus some job base classes in the subdirectory `Jobs`.

I have refactored (and tested) seven jobs into the new engine, so there are six less crontab entries after this change.
* publish caches (`run_publish`=> `PublishCachesJob`)
* new cache notification (`run_notify` => `NewCachesNotifyJob`)
* watchlist notification (`runwatch` => `WatchlistNotifyJob`)
* autoarchive caches (`auto_arch` => `AutoArchiveCachesJob`)
* fetch new geokrets (`geokrety.new` => `GeoKretyNewJob`)
* geokrety logging (`processGeoKretyDbQueue` => `GeoKretyDbQueueJob`)
* geopaths (`PowerTrailCronJob` => `GeoPathJob`)

The new OKAPI signalling job (#1769) is NOT refactored; it stays separate for performance reasons (slow external providers for the cache altitudes). So this PR does not affect #1769.

Also, I did not refactor `cache_titled_add.php`, which looks like a construction site.

The engine allows to
* disable single jobs per OC site (via settings.inc.php)
* run single jobs for testing (via util.sec/cron/run_cron.php + job name).

The intervals/times when the jobs are run are hard-coded; I tooke the times from `cron-job-list`. If different OC sites need different times, I can add config settings for that.

**Instructions for deploying this:**
* disable all the seven cronjobs listed above on ALL four OC sites
* merge this
* run the **sqlAlter**
* install the new cronjob engine, as shown in `cron-job-list`

(Alternatively, I could temporarily re-add the old cron job scripts. Then you could deploy this site by site; and when all four are done, the scripts can be removed again.)

Please priorize this as it fits your needs and time. It's not urgent for me in any way.


deg-pl:
@following5 Yours work is awesome!
BUT :)
1) We can't disable cronjobs on all 4 OC PL sites till merge. It is impossible. Please revert all old cronjobs - we'll deploy new cronjob engine site by site.
2) cache_titled_add.php on OC PL we run weekly, on OC UK and OC NL monthly, and on OC RO isn't run at all. Can we migrate this script into new cronjob engine, but with posibility to configure running interval per site?

deg-pl:
@following5 It is great! Very good job.
But... (please, don't kill me ;) ) settings.inc.php is our obsolete system. Cron engine config should be in /Config directory - see files included in this directory.
So... could you once more refactor your code?

kojoty:
> @following5 It is great! Very good job.
> But... (please, don't kill me ;) ) settings.inc.php is our obsolete system. Cron engine config should be in /Config directory - see files included in this directory.
> So... could you once more refactor your code?

I agree - we definitely would like one day to remove the "settings.inc.php" at all.

following5:
Nice. Have refactored the config. Please verify the schedule im Config/cronjobs.* for other sites than PL.

But how can I overload the settings now on my devel site? E.g. I want to disable the GeoKrety job:

`$cronjobs['schedule']['GeoKretyNewJob'] = 'disabled';`

following5:
When everything is ok, I would squash this all into one commit. Don't think the inbetween iterations are relevant in any way.

following5:
Have found a way to move the Okapi Signalling cronjob into the new cron engine. So _all 9_ OCPL cronjobs are now included here. Sorry for the duplicate work with updating the crontab ... I did not expect that this would work.

kojoty:
@deg-pl , @following5 

I don't have more comments for now - lets merge it. I assume that cron jobs are just copied - I don't try to compare the differences and the content of jobs.

I think it would be great to make a few minor changes/improvements - but let's do that later - not in one commit:

- removing util.sec/cron/run_cron.php at all - if we have a central cron class we don't need this - we can call that class directly - of course it needs small refactor to allowing call by local only

- trivial: obsoleted "jobs" scripts can be removed

I don't know if anyone of us check regular how the cron works so:
- it would be great to add simple way to check the cron results for "tech-admin" menu
- it would be great to add email alert sending on any error in cron 

Best regards from far...
kojoty

following5:
I have squashed the commits and fixed the Jobs paths.

> I assume that cron jobs are just copied

There are a few adjustments regarding Locks and error handling.

> it would be great to add email alert sending on any error in cron

This should already work, at least on my develsite it does. In case of error, a message is output to stderr or stdout. The cron daemon then sends an email to root.

following5:
Btw, when the new cron engine runs for the first time, it will not know the last run time of the jobs, and immediately run all jobs that may be overdue. From the second run, everything will run at the scheduled times.

deg-pl:
Is any way to overide this? For job which select cache of the week/month

following5:
It's okay for this job, because it is configured for a certain day of week or month. Will not run on any other day, and not before 21:55.

When should that job run on NL and UK sites? I configured it for day one at 21:55, i.e. next run on 1 January 21:55.

harrieklomp:
On OCNL we have 0  0  1  *  *
When i am correct first day of the month at 00:00
It can be changed later that night if needed.

following5:
Ok. I move it to 0:20, so there is better load distribution of cronjobs.

deg-pl:
@andrixnet We have new crontab engine made by following. To migrate to new engine switch off all current crontab jobs (except OKAPI cron5 and DB maintenance jobs) and add one job which should be done every 5 minutes - see /docs/crontabs/cron-job-list

@harrieklomp - I think you already use new engine ;)

harrieklomp:
@deg-pl Yes OCNL is running with new engine.

PR #1786, opened 2018-12-17
AutoArchive refactoring and new cronjob engine related changes

deg-pl:


deg-pl:
@andrixnet @harrieklomp There are some strings to translate. See autoArchive*

PR #1789, opened 2018-12-18
Proposal: public cronjob schedule

following5:
The cron schedule does not contain any sensitive information. It can by interesting for any developer and even for users when the jobs will run or did run.

So I suggest to make the cron admin page public at least for logged-in users, and only restrict _running_ jobs to sysAdmins.

### DIFF
--- a/Controllers/Cron/CronAdminController.php
+++ b/Controllers/Cron/CronAdminController.php
@@ -21,16 +21,16 @@ class CronAdminController extends BaseController
 
     public function index()
     {
-        $this->securityCheck();
-
-        $this->showCronAdmin();
+        if ($this->isUserLogged()) {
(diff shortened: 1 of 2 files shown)


deg-pl:
I think we don't need access users to cron jobs.

following5:
At least for developers that write new cronjobs it's useful. They can immediately check at all sites if the new job runs. (If it aborts due to an error, no "last run" date will be displayed).

As Okapi developer, I have learned to appreciate this kind of publishing non-sensitive sysinfo. E.g. without okapi/devel/dbstruct, I could not write the new DB updating tool. Without okapi/devel/attrlist, many Okapi development tasks would be very complicated.

And I can't see any disadvantages of that.

PR #1811, opened 2019-01-04
migrate APC to APCu

following5:
On the current devel VM there is no ACP installed - "Call to undefined function apc_fetch()". After changing to `acpu_fetch()` - which is already used in some other functions - everything works fine.

(In search-signatures.inc.php there is `acpu_store()` combined with `acp_fetch()` ... probably was forgotten to change [here](https://github.com/opencaching/opencaching-pl/commit/3a76c2f102d90bf0a3447c9f70ac3757117a323a#diff-2ad6aaa1ab10811bf9379fdee25978fc)?)

### DIFF
--- a/Utils/Cache/OcMemCache.php
+++ b/Utils/Cache/OcMemCache.php
@@ -8,7 +8,7 @@ use lib\Objects\OcConfig\OcConfig;
 
 
 /**
- * This is just wrapper for apc/apcu.
+ * This is just wrapper for apcu.
  *
  * PLEASE NOTE: that PDO object can't be serialized and every "oc-object" h
  * as reference PDO instance. It is handled in BaseObject class as
@@ -17,6 +17,7 @@ use lib\Objects\OcConfig\OcConfig;
  */
 class OcMemCache
 {
+    public static $debug = false;
 
     /**
      * Fetch from cache or create by callback-creator if there is no such entry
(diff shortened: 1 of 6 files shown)


following5:
Did you verify that it returned null instead of _false_? Or could it also have returned null instead of the value that was stored?

deg-pl:
@following5 No. I didn't verify it

following5:
Ok. There even is a third option: null was stored and then retrieved!

Have refactored this all into OcMemCache, and done two pro-forma fixes:
* return null if null was stored, not false (APCu *can* store null)
* return false from `requestSigner::extract_user` on bad input; though I suspect this is dead code (`get_signature_text` is not called anywhere)

following5:
> return null if null was stored, not false (APCu can store null)

More precise: Store null as placeholder '[OcMemCache null]', and return that as null when fetched. So the workaround for the assumed bug does not produce another bug (null => APCu => false).

kojoty:
BTW. 
I had an idea to used wrapper for APC code here: https://github.com/opencaching/opencaching-pl/blob/master/Utils/Cache/OcMemCache.php

why?
- to unify prefix which is same for all cache entries in OCPL code
- to be sure that noone is used apcu_entry - there were some problem with it 

so my proposition to refactor from APC to OcMemCache.

kojoty:
> > so my proposition to refactor from APC to OcMemCache.
> 
> That's what I did yesterday, see my commit and comment above.

Sorry for misunderstanding - I was offline all the weekend and it seems I overlooked this message :) 
Thanks

kojoty:
@following5, Thanks for the next great commit :)

PR #1814, opened 2019-01-08
refactored all $dateFormat-based date formatting onto Formatter::date()

following5:
All tested except for the change in badge.php.

In NewsAdminController.php and Converter.php, only line endings are changed (CRLF => LF). Did some code changes there and discarded them.


following5:
$dateFormat then may be moved to new Config; there are no references left outside of OcConfig class.

kojoty:
Thanks :)

Issue #702, opened 2016-06-30
COG stats stopped working on pl node


mzylowski:
After package upgrade by @deg-pl on PL node COG stats doesn't work.

![13552696_1295608723801377_1485786475_n](https://cloud.githubusercontent.com/assets/13780868/16499774/1e4b2bec-3f02-11e6-9c65-0d5c424ada1f.png)

> Empty input data array specified for plot. Must have at least one data point

And from logs

>  [Thu Jun 30 20:05:56.934661 2016] [:error] [pid 30255] [client 89.65.18.106:51456] PHP Warning:  max(): Array must contain at least one element in /srv/ocpl/graphs/COGstat.php on line 92, referer: http://opencaching.pl/articles.php?page=cog
[Thu Jun 30 20:05:56.934731 2016] [:error] [pid 30255] [client 89.65.18.106:51456] PHP Warning:  max(): Array must contain at least one element in /srv/ocpl/graphs/COGstat.php on line 92, referer: http://opencaching.pl/articles.php?page=cog

deg-pl:
It is not related with package update. It is related with:
```
# Setup begining of stat for OC Team. Start timie 01 June every year
    $year = date('Y');
    $year_old = $year - 1;
    $year_new = $year + 1;
    $count_days = date('z');
    if ($count_days < 181) {
        $start_time = $year_old . '-07-1 00:00:00';
        $title3 = iconv('UTF-8', 'ASCII//TRANSLIT', tr('number_month')) . ' ' .$
    } else {
        $start_time = $year . '-07-1 00:00:00';
        $title3 = iconv('UTF-8', 'ASCII//TRANSLIT', tr('number_month')) . ' ' .$
    }
```

following5:
The issue with Jgraph error messages on empty COG stats has been fixed.

>  mzylowski commented on 30 Jun **2016**

In leap years, this ...
> if ($count_days < 181) {

should be 182. Needs fix before 30 June 2020 :-D

PR #1820, opened 2019-01-11
fixed handling of empty COGstats

following5:
Quick fix for issue #702. Can be improved by showing individual placeholders for each empty graph, and by making the "no data" message translatable.


kojoty:
BTW: merged in Friday because:
- this is change in COG panel only
- we don't want more error-emails

Issue #723, opened 2016-08-07
Hardcoded provinces in statistics


kojoty:
from @harrieklomp message:

> It could be interesting for other nodes if the Polish provinces are not hardcoded in the software like in http://www.opencaching.nl/articles.php?page=s7 or http://www.opencaching.nl/articles.php?page=s8 . There should be a system that nodes can use their own provinces.

It should be fixed as well...

harrieklomp:
:+1:

deg-pl:
Fixed by commit f5aee81da300f3dee63d0d8ab8248dc634379550 . Thanks @following5

PR #1824, opened 2019-01-13
enable s7 and s8 statistics for other countries than PL

following5:


deg-pl:
@following5 does this fix #723 ?

following5:
yes

Issue #600, opened 2016-05-04
Node cannot exclude cache size


harrieklomp:
In the recent past when cache sizes where read via database a node could decide to exclude a cache size. This was done via settings.inc.php.

```
'forbiddenCacheSizes' => array(
    cache::SIZE_MICRO
),
```

Is it possible to make this work again?

harrieklomp:
The first post is not correct!!!! 

When i remember correct i did add this myself as prework to exclude cachesizes. This was partly for OC.RO who liked to add a new cachesize "nano" and partly to prevent making caches like "nano" and "micro" caches on OC.NL. I do not have knowledge to implement this.
On the live site (oc.nl) i did edit /lib/cache.php to prevent selecting size micro. But somehow this was overwritten by github.

This means that i need to change my question.
Can a programmer make it work that we can exclude cachesizes?

kojoty:
@harrieklomp,
I'm going to redesign configuration ( #290 ) and after that also move the configuration which was stored in db tables to config files. I'll try to resolve it then - then every node will be able to define its cache sizes.

andrixnet:
> every node will be able to define its cache sizes.
Meaning a common, complete set of sizes defined project-wide and each node configures which items to have active (and in use) and which not (for new caches and edit caches)

following5:
> Can a programmer make it work that we can exclude cachesizes?

On my develsite, the setting

`$config['forbiddenCacheSizes'] = [cache::SIZE_MICRO]`

works.

harrieklomp:
This is solved now.

Issue #876, opened 2017-02-24
cache_size data definition and properties


andrixnet:
Cache size defines the size of the container for a cache. Each cache size is identified by a unique ID.

Properties:
- ID
- name (for referencing translations)
- base image icon
- actual (represents an actual size; eg false = no container, not specified)

andrixnet:
Initial corrections done here: https://github.com/opencaching/opencaching-pl/projects/2
NANO and NOT_SPECIFIED cache sizes that already existed in the code as non-working and incomplete are now fully functional.

andrixnet:
DOcumented at https://wiki.opencaching.eu/index.php/Cache_sizes

andrixnet:
Documented at https://wiki.opencaching.eu/index.php/Cache_sizes
@kojoty plese take a look at the above link.

andrixnet:
@harrieklomp @kojoty @deg-pl I have an important functional question (might have been partially discussed elsewhere, but needs clarifing here): 

there are several cache sizes. 
size_A is forbidden. 

The question is what to do in the following scenarios:
- new caches of size_A are NOT ALLOWED _(this is clear)_

- existing caches of size_A, when edited, what to do?
1. size_A is not allowed, automatically reassign the size upon submission of cache edit form (to whatever the first option is in the dropdown. (assume user does not change it, is unaware of it))
2. all caches of size_A, once it is forbidden, may be allowed to stay as such, when editing, other caches may not set size_A.
3. admin changes database and reassigns all caches of size_A to the closest available size. (assume site rules define the allowed sizes, conversion and give the admin this power)

if size_A is forbidden, but there are older caches of size_A, allow the following:
- search for caches with size_A ? (general search)
- search caches in my routes with size_A ? 

There is also the consideration expressed by some, that allowing users to see a searchable cache size, that is not allowed for new caches, creates discussions, raises questions and requests to allow that cache size to exist for new hides (even though OC rules do not allow it and it is deemed undesirable).

kojoty:
My vision is simple and the same for sizes/types/etc.

A/ There is a wide super-set of some entries across the code common for all nodes
B/ Node define in settings subset of allowed-entries (entries which can be found at this node) 
C/ Node define in settings subset of forbidden-for-new-caches entries (entries are present, but can't be set for new caches)

Then everywhere (except new cache dialog) are presented entries form subset B and additionally on cache edit there is a check which prevent change any entry from subset B to any from subset C.

I hope this is clear.

[...]

PR #1837, opened 2019-01-16
enabled nano cache size

### LINKED ISSUE #1835: Introduce cache_size Nano
OC.NL has re-introduced the size Micro again.
To go a step further than only show micro (like GC) it should be nice to also show Nano caches in the system.

As suggestion the Nano could have ID 1 because the Micro has ID 2. The bigger the caches sizes, the biggest the ID number.

Because the search page looks at the normal DB this can easy be solved by importing the table "cache_size" that can be uploaded in /sqlAlters. The search page would loo like:
selectie_289

Is it possible to add this Nano size as ID 1 to the system?

following5:
This fixes #1835.

* Cache size "Nano" is enabled for all OC sites. If some site does not want it, it can be disabled in Config/site.XX.php.
* Cache size "Not specified" is enabled only for OC RO, where some of those caches exist. https://github.com/opencaching/okapi/issues/519 should be fixed asap.
* All size search and filter pages adjust to the existing caches. E.g. "Nano" search option will appear when the first Nano cache is published on the site.
* Table `cache_size` becomes obsolete.
* Fixed some PHP notices on 'mobile' pages.


deg-pl:
Please disable nano size by default on all OC sites, and enable only for OC NL.

lib/Objects/GeoCache/GeoCache.php
```diff
@@ -1,6 +1,7 @@
 <?php
 namespace lib\Objects\GeoCache;
 
+use Utils\Database\OcDb;
```
deg-pl: @following5 You have acces to OcDb via $this->db (from BaseObject). So don't use OcDb this way please.

following5: I could use explicit `\lib\Objects\OcConfig\OcDb::instance()` to get rid of the use statement. But there are lots of other static GeoCache:: methods that currently use XDb and will reference OcDb class when refactored to OcDb.

deg-pl: Possible in static method via self::db() (see BaseObject)

following5:
This change also fixes issues #600 and #876.

PR #1839, opened 2019-01-17
1831 in context localization

kojoty:
Hi, this is the next "too big refactor" but I got the flow when I tried to introduce "Crowdin-in-context mode".

What was done:
- **Crowdin-in-context mode can be activated and deactivated by link near the flags in the footer**
- **new configs: `/Config/I18n.*.php`**
- complete refactoring of language.inc.php => I18n class
- reporting of missing translations (requested phrases missing even in en.php)
- global var $lang is now removed across the whole code
- many cleanups in context of language is done

What still needs to be done:
I would like to review all the places when we use "translation-in-db" and refactor to use new method: 
`I18n::getLangForDbTranslations()` - the idea is to use "default" lang if there is no requested lang - it allow to use all languages without "complicated" sync the DB - of course finally we should get rid of translations in DB.

### DIFF
--- a/okapi_settings.php
+++ b/okapi_settings.php
@@ -44,7 +44,7 @@ function get_okapi_settings()
         'DB_NAME' => $dbname,
         'DB_USERNAME' => $dbusername,
         'DB_PASSWORD' => $dbpasswd,
-        'SITELANG' => $lang,
+        'SITELANG' => 'en', //TODO: how to read it from I18n class?
         'SITELANGS' => array_map('strtolower', $config['defaultLanguageList']),
         'SITE_URL' => isset($OKAPI_server_URI) ? $OKAPI_server_URI : $absolute_server_URI,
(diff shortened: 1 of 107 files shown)


okapi_settings.php
```diff
@@ -44,7 +44,7 @@ function get_okapi_settings()
         'DB_NAME' => $dbname,
         'DB_USERNAME' => $dbusername,
         'DB_PASSWORD' => $dbpasswd,
-        'SITELANG' => $lang,
+        'SITELANG' => 'en', //TODO: how to read it from I18n class?
```
kojoty: @following5 I wonder how to deal with this - now config is accessible by class call - not by global vars... Maybe you have some idea...

following5: Made a proposal here: https://github.com/kojoty/opencaching-pl/pull/2

PR #1846, opened 2019-01-18
fast fix for email flood

kojoty:


kojoty:
see #1847 for details of the problem

Issue #1244, opened 2017-10-03
Contradictions in bicycle attribute


following5:
@harrieklomp @andrixnet 
At OC.nl and OC.ro there are contradicting definitions of the "bicycle" attribute.

OC.nl:
* English and German: "Bicycle recommended"
* Dutch and Polish "Can be reached by bicycle"

OC.ro:
* English "Bike"
* Romanian: "Bicycle recommended"
* Dush and Polish: "Can be reached by bicycle"

I think you should decide for one of both meanings. :-)

At OC.pl it is:
* English "Bike"
* Polish: "Can be reached by bicycle"

harrieklomp:
Perhaps there could be made a standard English translation for all available attributes. This should include all attributes from all OC nodes.
This could be placed on https://opencaching.eu/index.php/Cache_attributes or somewhere else. Each node can add their own translation.
These translation can be used for databases and OKAPI

This is only a suggestion ;-)

following5:
We have standard translations for all attributes [here](https://github.com/opencaching/okapi/blob/master/okapi/services/attrs/attribute-definitions.xml) in Okapi. But those may be wrong for the NL and RO "bicycle" attribute.

You must to decide what the meaning of this attribute is on your OC site. If it is "bicycle recommended", then a new Okapi attribute must be defined for that, which will have other meaning and translations than the PL attribute.

harrieklomp:
If English translation and meaning in OKAPI is clear than all other translation should be the same.
I can only suggest (personal opinion) to make a standard English text and every node should translate it like it is. We can make a useless discussions about the bike attribute or any other attribute but we must make a standard.
If all English text in [OKAPI ](https://github.com/opencaching/okapi/blob/master/okapi/services/attrs/attribute-definitions.xml)are correct i will translate these in Dutch. I will use the same translations in database.

harrieklomp:
Does any other node OCPL, OCUK, OCRO and (perhaps) OCUS agree with this?

following5:
> If English translation and meaning in OKAPI is clear 

The meaning in Okapi is not clear for the NL and RO "bicycle" attributes. The may have been wrongly added to A27.

Please decide what is the _meaning_ of this attribute at OC.nl. Does it mean "accessible by bycycle" (likt at OC.pl), or "bicycle recommended"? These are different things.

following5:
The bicycle attribute is the only one with this issue; all other are fine.

[...]

Issue #806, opened 2016-11-01
Attribute "Recommended for kids"


andrixnet:
Attribute "Recommended for kids" is listed as having id=41.
According to OKAPI, OCDE has the same attribute with id=59 (unused by OCPL-current)
The result is two A code mappings in OKAPI for the same attribute. 

With the entry of OCUK to OCPL-current, OCUK brings many attributes that differ from OCPL-current, some similar to OCDE, some completely new.
OCUK already has attribute "Not available at high tide" id=41. Knowing the overlap (thanks to OKAPI), this can be easily resolved quickly and elegantly: 

According to http://opencaching.pl//okapi/devel/attrlist and http://www.opencaching.nl/okapi/devel/attrlist there is no attribute id=59 in use. 

OCPL and OCNL:
Run the following queries on your servers (and in devel machines) :  
```
update cache_attrib set id=59 where id=41;  
update caches_attributes set attrib_id=59 where attrib_id=41;
```

andrixnet:
OKAPI fix requested by https://github.com/opencaching/okapi/issues/445

andrixnet:
There are more attribute id duplications and overlaps, but for the time being, these area the easies to solve. (and among the ones with priority for integrating OCUK properly)

deg-pl:
I think is MUCH easier to change one site - OC DE, not 4 different sites (NL, RO, PL, UK)

andrixnet:
I can handle RO, help Harrie with NL, you can handle UK and (probably) PL. 
Anyway, AFAIK things seem to happen much faster at OCPL development then OCDE.

deg-pl:
I'm afraid malfunctioning OKAPI during changes. Is it possible to include dual attributes (with old and new ID)?
@wrygiel What do you think about it?

andrixnet:
DO you mean the same ID fo rhte same OC site written in 2 A codes?

deg-pl:
No. One A-code assigned to two ID (new and old). But it was a question. I don't know that it is possible. I don't know OKAPI well.

wrygiel:
> I'm afraid malfunctioning OKAPI during changes.

I will be leaving for holidays very soon, and I won't be able to help you much! However, you shouldn't be too afraid - only a couple of apps make use of attributes, so even if you break something, the impact will be minimal.

> No. One A-code assigned to two ID (new and old). But it was a question. I don't know that it is possible.

Yes, it's okay to assign many internal IDs to a single A-code.

[...]

PR #1852, opened 2019-01-19
Add NL,UK and RO attributes to en.php

following5:
This is a proposal for a unified OCPL attribute set. These `attribute_`  translation strings are not in use yet, but will be needed soon.

The proposal includes some clarifications for attributes wich have ambivalent meanings in the current `cache_attrib` table texts:

* 44: **Wheelchair accessible**. This is also called "Accessible for disabled", but has a wheelchair symbol. "Disabled" could also mean blind or mentally disabled or whatever, which makes a big difference.
* 48: **Bring your own pen**. This is also called "Take something to write", which may mean to bring pen and paper for making notes, but it has a Pen symbol. I guess the intended meaning is BYOP - there is no pen in the cache. Right?
* 80: **Only at certain times of day or week**: See https://github.com/opencaching/opencaching-pl/issues/806#issuecomment-455805256.
* 84: **Access only by walking**. This is also called "Access by walk", but "only by walk" makes more sense, because most caches are accessible by walk.

Also, I have combined the attribute 61 names "Monument", "Monumental Place" and "Historic place" to "Monument / Historic place".

@andrixnet @harrieklomp If you don't agree to my numbering proposal and still prefer your (different) proposals of completely renumbering all attributes, please close this PR for further discussion. We have to postpone some other issues then, which are blocked by issue  #1244. Especially to give all OC sites the option to enable any of the translated languages.


following5:
Please note that the *display order* of the attributes will be made configurable independent from the *ID order*. So the order of the IDs is not important.

harrieklomp:
> @andrixnet @harrieklomp If you don't agree to my numbering proposal and still prefer your (different) proposals of completely renumbering all attributes, please close this PR for further discussion. We have to postpone some other issues then, which are blocked by issue #1244. Especially to give all OC sites the option to enable any of the translated languages.

I am not that difficult, just do what is needed :-)

following5:
Thanx :)

PR #1883, opened 2019-01-23
merge cache_logs triggers

following5:
Old OC NL <s>MySQL</s> MariaDB cannot define multiple triggers on the same table with the same action.

Workaround: Merge `cl_insert` trigger to `cache_logs_insert` and `cl_update` to `cache_logs_update`.


following5:
Just noticed that 100_init already was run three days ago at OC UK (maybe also at OC PL?) So this change won't work there. Will need a new update.

Issue #1885, opened 2019-01-23
Logs not working properly at OCRO


andrixnet:
Recently I discovered that logs are not shown for any cache, neither the viewlogs page works. 
In cache page there are no autoloaded logs. 
Separetely, viewlogs.php shows an empty page. 

Server log shows: 
```
PHP Parse error:  syntax error, unexpected '->' (T_OBJECT_OPERATOR) in /wwwroot/viewlogs.php on line 203, referer: https://www.opencaching.ro/viewcache.php?cacheid=133
```

following5:
What PHP version is running on OC RO?

following5:
Problem solved, OC RO logs are back. :-)

PR #1886, opened 2019-01-23
syntax fix for older PHP versions; updates #1885

following5:
I am not sure if I may directly merge fixes like this, or if kojoty or deg-pl want to review it ...?

### DIFF
--- a/getLogEntries.php
+++ b/getLogEntries.php
@@ -65,7 +65,7 @@ $logEntryController = new LogEntryController();
 $logEntries = $logEntryController->loadLogsFromDb($geocacheId, $includeDeletedLogs, $offset, $limit);
 $result = '';
 
-$logfilterConfig = (OcConfig::instance())->getLogfilterConfig();
+$logfilterConfig = OcConfig::instance()->getLogfilterConfig();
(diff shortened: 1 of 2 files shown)


kojoty:
@following5 why do you think that you can't merge things like this?

My general thinking about cooperation here is like that:
- if I have a emergency fix for something what breaks the service - commit ASAP
- We establish the common rules that we in generally should use PR for code deliver + not to merge PRs during the weekends (free days in generally) because this is a time when much more people use our site and we want to make it stable
- if I have a fix for something - but this is not very important - I waiting more or less but I want to show others what I did
- if I have a enhancement - something what is not urgent - I create PR and wait a day or two (sometimes more) for comments
- I try to inform others if I made or found something in my opinion important what other developers should be aware
- I try to discuss ideas, changes and problems which are worth it :)

I have full confidence in your competences so please feel free to continue your great work on OCPL code!

..and yes - I like to take a look in other developers changes - but I can do that also after PR is merged.

following5:
@kojoty 
> why do you think that you can't merge things like this?

Because of my experiences with OCDE project, which enforces a hierarchical organization and is very restrictive on commit rights. Glad to see that OCPL does it more common-sense like OKAPI. :-)

Issue #1693, opened 2018-10-11
Admin options on GeoPaths


deg-pl:
Now admins (OC Team / COG in OC PL) can't see deleted GeoPath's comments and can't restore them. This should work like cache logs - OC Team see deleted logs as strikeout and can restore them.

following5:
Can someone please test this on a production site?

harrieklomp:
A small test on OCNL
After click on "Restore" the log is still strikeout and not restored. We get a small error like this:
```
Error: Using $this when not in object context



#0 /serve/www/oc_www/oc_www/powerTrail/ajaxRemoveOrRestoreComment.php(28): lib\Objects\ApplicationContainer::getLoggedUser()

#1 {main}
```

harrieklomp:
BTW. The mail does not have the nodes own logo.
![selectie_295](https://user-images.githubusercontent.com/1060518/51683627-9559a800-1fea-11e9-9466-221a640c0135.png)

following5:
Have applied a fix for restoring Geopath comment; can you test again?

harrieklomp:
Tested again and now it works. Log is restored :-)

following5:
Now there is the node's logo in Geopath comment notification emails.

PR #1890, opened 2019-01-24
use configured logo in geopath-comment notification emails; updates #1693

following5:
Note: I believe that the &lt;img> "alt" attribute should only be supplied if the image contains essential data. E.g. consider a blind user using a screenreader. The screenreader would say "Opencaching Logo" here, which is not necessary.


deg-pl:
@following5 img alt atribute is REQUIRED in all img tags. May be empty (alt="") but must exist.

PR #1891, opened 2019-01-24
fixed Admin.DbUpdate bug introduced with aaf82371

following5:


following5:
Direct merge, because not in production use yet, admin-only page, and need it for further development.

PR #1892, opened 2019-01-25
Db update deployment test1

following5:
This is a test for automatic DB updating.

1. Add the DB updater (util.sec/dbupdate.php) to the [deployment scripts](https://github.com/opencaching/opencaching-pl/tree/master/docs/system) at OC PL, UK and NL.
2. Merge this change.

"101_deployment_test_1" then should run automatically, so the Admin.DbUpdate page should show that it did run. If not, please DO NOT run it manually. Instead, we must fix the problem and then test again.

### DIFF
--- a/Utils/Database/Updates/101_deployment_test_1.php
+++ b/Utils/Database/Updates/101_deployment_test_1.php
@@ -1,0 +1,26 @@
+<?php
+
+// created on 2019-01-25 by following5
+
+namespace Utils\Database\Updates;
+
+return new class extends UpdateScript
+{
+    public function getProperties()
+    {
+        return [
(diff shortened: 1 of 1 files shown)


following5:
@deg-pl @kojoty 
How is the progress with enabling the DB update system at OC PL and UK? This blocks many open issues, which need DB changes.

@harrieklomp 
I can help with installing this at OC NL, if someone knows where the deployment script is located.

following5:
There is a file access rights issue when manually running util.sec/dbupdate.php. Also, this should be moved to the new `/bin` directory before installing.

following5:
Testing OC NL database update

andrixnet:
Done at RO.

PR #1894, opened 2019-01-26
remove default country PL

following5:
Don't think there is something like a "default country". The setting for OC PL already is in site.pl.php.

### DIFF
--- a/Config/site.default.php
+++ b/Config/site.default.php
@@ -3,28 +3,16 @@
 use lib\Objects\GeoCache\GeoCacheCommons;
 
 /**
- * This is simple configuration of maps in the OC code
- *
- * This is a DEFAULT configuration for ALL nodes, which contains necessary vars.
- *
- * If you to customize it for your node
- * create config for your node and there override array values as needed.
+ * DEFAULT general site properties for ALL nodes
  */
 
-/**
- * Configuration of general site properties in defaut version
- */
-
-/**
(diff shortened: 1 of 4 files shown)


kojoty:
@following5 
This is detailed case but I write it:
my general idea of default config is to ensure that site with only "defaults" is able to work - then node admin can configure/customize what is needed 

I'm not sure if set by default not accepted value and make user to change it by exception is not (in my opinion) the best solution - what're your thoughts?

kojoty:
right now we have a mess in config - most of config vars are still in old settings.* files - a few new settings is refactored to new files

I think we need to continue refactor and one day we also clean up the config files. 
I believe noone needs to set up oc node without customization but maybe at the beginning let's not make the new admin to configure more than necessary.

Maybe it should be better to set EN instead of PL - PL was only because I thought also about dev sites - right now most OCPL code devs are from PL - but this is nothing important

following5:
There is no country "EN". :-)

Agree that the default language may be set to 'en'.

following5:
We could allow to set this in settings.inc.php, if it is not configured in Config/site.XX.php. Then

* it is still easy to set up a site (no code change needed),
* the site can be set up so that it fully works,
* the authoritative settings stays in Config/site - it cannot be overridden locally.

following5:
What about using the TLD of the site's URL as country default?

following5:
will make a new proposal

PR #1895, opened 2019-01-26
Add automatic DB trigger and procedure updating 

following5:
See [documentation](https://github.com/following5/opencaching-pl/blob/db-routines/docs/DbUpdate.md#user-content-db-routines-triggers-procedures-and-functions) and the Admin.DbUpdate page, which now includes the .sql files with trigger/proc definitions. All the triggers and procs are tested.

Utils/Database is not the perfect place to put all that database definition and update code, but I did not find a better place. Maybe this can be decided later when other directories like 'lib' are refactored?

This change can be merged independently from the status of DB-Update deployment.


Utils/Database/DbUpdates.php
```diff
@@ -184,4 +217,69 @@ public static function getUpdatesDir()
...
+
+    public static function runRoutines($fileName)
+    {
+        $queries = self::getRoutineContents($fileName);
+        if (substr($queries, 0, 10) != 'DELIMITER ') {
+            throw new Exception('DELIMITER statement is missing in '.$fileName);
+        }
+        OcDb::instance(OcDb::ADMIN_ACCESS)->simpleQueries($queries);
+
+        Facade::cache_set(
```
following5: The OKAPI depedency here is ok for now, but may be annoying when setting up a new OC site. May eventually create circular dependency between setting up OKAPI and OCPL site. 

The information should be stored in OCPL database instead, but that will need a DB update, which is not available yet at all sites.

following5:
Ok, let's go. I reviewed and tested all the DB routines in detail, so there is a good chance that they are bug-free.

PR #1900, opened 2019-01-27
direct update of watchers count in listing

following5:
Increment / decrement the displayed watchers count when watching / unwatching a cache.

### DIFF
--- a/Controllers/UserWatchedCachesController.php
+++ b/Controllers/UserWatchedCachesController.php
@@ -185,7 +185,8 @@ class UserWatchedCachesController extends BaseController
             $this->loggedUser->getUserId(), $cacheWp);
 
         if($resp){
-            $this->ajaxSuccessResponse();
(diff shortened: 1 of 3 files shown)


Controllers/UserWatchedCachesController.php
```diff
@@ -208,7 +209,8 @@ public function addToWatchesAjax($cacheWp)
             $this->loggedUser->getUserId(), $cacheWp);
 
         if($resp){
-            $this->ajaxSuccessResponse();
+            $cache = new GeoCache(['cacheWp' => $cacheWp]);
```
kojoty: @following5 
BTW: in my opinion much more cleaner way to create geocache object is "factory" method:
like that:
```$cache = Geocache::fromWayPointFactory($waypoint)```
(sometimes I think we even can create separate GeocacheFactory class, but for now this is like that)
what do you think?

Issue #1503, opened 2018-04-18
Incorrect GPX export while cache has ampersand in GC waypoint


deg-pl:
If cache has ampersand in GC waypoint:
![image](https://user-images.githubusercontent.com/13884863/38955229-4d9f07b2-4354-11e8-9653-6d546d55e6cc.png)
exported GPX is not valid XML:
`<oc:other_code>GC7GH&E</oc:other_code>`

Todo
- [ ] validation format of additional waypoints while user edits cache
- [ ] properly escape ampersand sign in GPX export

andrixnet:
Validation is a must. 
But is & really a valid character in waypoint code?

Btw, http://coord.info/GC7GH&E yields geocaching.com 404 page.

deg-pl:
No - valid GC waypoint in this case is GC7GH7E

andrixnet:
Then in this case it is a user error typing the code. 
IMO data validation should be done on cache edit page to ensure data is entered correctly, then GPX (and other exports) will work properly.

following5:
> properly escape ampersand sign in GPX export

Better remove invalid characters. GC codes must only consist of 0-9 A-H JK MN PQR T V-Z

These may be mapped:
* I => J
* L => J
* O => 0
* S => 5
* U => V

following5:
Bad GC waypoints now are omitted from GPX.

PR #1903, opened 2019-01-28
validate waypoint codes; updates #1503

following5:


lib/search.gpxgc.inc.php
```diff
@@ -490,9 +490,11 @@ function getPictures($cacheid, $picturescount)
...
 
         $other_codes = [];
-        foreach (['GC', 'TC', 'NC', 'GE'] as $platform) {
-            $code = strtoupper($r['wp_'.strtolower($platform)]);
-            if (substr($code, 0, 2) == $platform && strlen($code) >= 4) {
+        foreach (['gc', 'tc', 'nc', 'ge'] as $platform) {
+            $code = $r['wp_'.$platform];
+            $validatorMethod = $platform.'Waypoint';
+            $code = call_user_func(['\Utils\Text\Validator', $validatorMethod], $code);
+            if ($code) {
```
following5: This is the fix for #1503.

following5:
Let's go. This will force users to fix bad old waypoint code entries when editing caches. For example, there are ~50 active caches with OX waypoints (dead Opencaching.com) entered as Terracaching WP.

Issue #1608, opened 2018-08-31
Titled caches adding minor remarks


rapotek:
Some remarks of rather minor importance to `cache_titled_add.php`:

- when `cache_titled` table is empty, the next titled cache will never be generated for "week" or "month" periods, because `dataMax` retrieved from database will be null and therefore `DateTime` object (`dStart`) created on its base will always be current datetime,
- when there is no cache selected as a next titled cache, the script will throw an error (with an email sent) on 'insert' multiValueQuery,
- the script itself, although secured in some aspects, should not be available for public requests anyway

PR #1904, opened 2019-01-28
fixed and refactord titledcachesadd; updates #1608

following5:
* refactord TitledCacheAdd cronjob (mostly code style)
* fixed minor issues, see #1608 
* added minor safety check to CronAdmin against acidentally re-running of jobs that do not allow it


following5:
> added minor safety check to CronAdmin against acidentally re-running of jobs that do not allow it

This change is a bit too strict - some manual override should be possible for testing/debugging, and for the case that the first run partially failed. Will change that later today.

PR #1910, opened 2019-01-29
DOC: Folders structure draft

kojoty:
@deg-pl @following5 @harrieklomp @andrixnet 

I tried to describe my proposition of the final folders struct - next try :)
Please take a look! Feel free to comment and add fixes.


docs/OCPL_folders_structure.md
```diff
@@ -0,0 +1,127 @@
...
+		- .htaccess
+		- .gitignore
+	- **migration**: all scripts from here will be temporary migrated to `/public`, but finally should be refactored to controllers  
+
+### `/bin/` 
+	- **access**: NO access for webserver (PHP not allowed - only command line)
+	- root-level directory for command-line executable files
+	- **migration:**  contains scripts which are used ONLY from command-line - now OCPL has such in :
+		- `/ci`
+		- `/util.sec`
```
deg-pl: I run scripts from util.sec not only by php cli, but also via web page

docs/OCPL_folders_structure.md
```diff
@@ -0,0 +1,127 @@
...
+		- hosted JS libs (like tinyMCE)
+		- CSSes
+		- other STATIC files
+	- folder structure - see below 
+	- **migration** - see below
+
+### `/resources/`
+	- **access**: RW for webserver, (TODO: PHP ?)
+	- a root-level directory for files which needs webserver RW access for uploaded files etc.
+	- let's store main directory structure in git, but the content should be ignored (by git-ignore)
```
deg-pl: gitignore should be configured to ignore files, but not directiories. We should store in repo directory schema

docs/OCPL_folders_structure.md
```diff
@@ -0,0 +1,127 @@
...
+### `/public/css/`
+	- **contains** CSS "common" files used by many views
+	- TODO: do we need to store anything here ?!
+
+### `/public/img/`
+	- **contains** common used images, icons etc.
+	- images should be grouped in logical structure of folders
+	- every folder with images should contains README file with description of origin/author of images
+
+## 5. `/srv/` folder structure
```
deg-pl: srv? or src?

docs/OCPL_folders_structure.md
```diff
@@ -0,0 +1,127 @@
...
+	these images can be stored here as well
+
+### `/public/js/libs/`
+	- **contains** JS external libs which are hosted in OCPL code (like tinyMCE)
+	- every lib has its own folder + README file with description how to update this lib
+	- JS shoudl be loaded to Views in OCPL code by ONE chunk (every lib has only one reference in PHP code)
+
+### `/public/js/`
+	- **contains** JS "common" scripts used by many views - but not external libs (see above)
+	- TODO: do we need to store anything here ?!
```
deg-pl: yes. for example cookiesInfo.js

docs/OCPL_folders_structure.md
```diff
@@ -0,0 +1,127 @@
...
+
+### `/public/css/`
+	- **contains** CSS "common" files used by many views
+	- TODO: do we need to store anything here ?!
```
deg-pl: yes. Main, common CSS's ;)

deg-pl:
A very good idea. I've written some comments. Thanks!

docs/OCPL_folders_structure.md
kojoty: Scripts which are running from WEB should (in my opinion) be refactored to MVC model and we should have a controller for such scripts. But in util.sec are also scripts which are run only from CLI - and such scripts should be moved to /bin

docs/OCPL_folders_structure.md
kojoty: right!

docs/OCPL_folders_structure.md
kojoty: no? If main.tpl.php is for example in `/view/common` or something like that I believe main css shoudl be also in `/public/view/common/main.css` - my idea is to reflect `/srv/View` structure in `/public/view` structure 

What do you think?

docs/OCPL_folders_structure.md
kojoty: this is typo

docs/OCPL_folders_structure.md
deg-pl: ok. Now I understand. You're right!

docs/OCPL_folders_structure.md
deg-pl: ok

following5:
Some opinions, in alphabetical order:

**/logbook/**

=> /public/logbook/ - if it can be repaired.

**/mobile/**

Don't think it fits anywhere else. And it is a website of its own, which justifies a toplevel directory.

**/okapi/**

=> /vendor/oc/okapi/

**/src/**

Under this name, I would expect all nonpublic PHP source code of the website including /config and /i18n.

**/resource/**

Strongly prefer **/var/**.

Regarding the substructure, I would prefer to maintain that dynamically by PHP code. Storing it in Git would be redundant information, because the directory paths are also string constants in PHP code. And redundant information tends to become unmaintained and outdated (see the failed /docs/sql experiment). The .gitignore thing also can become complicated.

Would rather implement some mechanism in the code that maintains the /var/ substructure, with each directory path existing only as ONE string constant.

**/vendor/**

Ok, standard directory for third-party libraries.

**/xml/**

=> /public/xml/

kojoty:
> > **/src/**
> > Under this name, I would expect all nonpublic PHP source code of the website including /config and /i18n.
> 
> right!

Thanks @deg-pl - I agree `/config` should stay in `/`

following5:
Agree, after some more thoughts. The config and language files are formally PHP source code, but logically not. They could have any other format than .php.

> /resource/ is a name from skeleton https://github.com/php-pds/skeleton - that's why I would use this name - but I'm not radical in this opinion.

"resource" usually means read-only data. A [resource](https://en.wikipedia.org/wiki/Resource) is something original that you fetch and process. E.g. geographical data to initialize the `gns_` and `nuts_` tables would be resources. Maybe the SQL trigger & proc files also can go there?

* /resource/sql/routines

It's not a perfect fit, because they are no data but source code. But they don't fit into any other of those toplevel dirs.

following5:
Generally, I would prefer to start the new structure in a clean way, not with exceptions and hacks, that create new "legacy" that will need another cleanup in the future. Experience tells that the final cleanup will never be done, if the dirt has been swept under the carpet (i.e. hidden in some obscure subdirectories).

* Create the new toplevel dirs
* Step by step move things there that fit
* Refactor things that don't fit *before* they are moved

Forcing everything into the new structure would just pretend that it is clean and hide the real problems that must be tackled. That would definitely not comply to the "skeleton model".

Regarding OKAPI: It is distributable by composer (OCDE does that). But agree that it may be better to not put it into /vendor, as long as it is not really distributed by composer. Also, the somewhat unconventional handling of composer in OCPL project (store copies of the checked-out packages in the Git repo!) makes it unattractive to move OKAPI deployment to composer.

Putting OKAPI into /public/ would be a huge break of the "skeleton" model. Most files there are not public.

Having said all that: I can handle any kind of messy structure, so if you prefer to replace the big mess by a smaller mess, please just do it. :)

following5:
> Having said all that: I can handle any kind of messy structure, so if you prefer to replace the big mess by a smaller mess, please just do it. :)

Sorry for the provocative wording. Better: Whatever is decided, I will try to help with the cleanup.

kojoty:
> Regarding OKAPI: It is distributable by composer (OCDE does that). But agree that it may be better to not put it into /vendor, as long as it is not really distributed by composer. Also, the somewhat unconventional handling of composer in OCPL project (store copies of the checked-out packages in the Git repo!) makes it unattractive to move OKAPI deployment to composer.

Yeah - our way of use composer is atypical - but in my opinion it fits best now - if we want to be sure that all nodes has same code - with same version of ext. libs we have two ways:
- store ext.libs. code in repo - what exactly we do now
- run composer update as part of update process and handle situation when it for example fails... 

@following5 How it should work in your opinion?


> Putting OKAPI into /public/ would be a huge break of the "skeleton" model. Most files there are not public.

@following5 I'm open for your suggestion - where to put OKAPI and be sure everything works OK - finally it would be nice to switch "documentRoot" to `/public` (some day)

kojoty:
I have a one more idea for `/resources` and ocpl-dynamic:
Let's add `/public/var` and store uploaded files there - in fact uploaded (or dynamically created content) is accessed often statically - so it should be in `/public`  - what do you think?

kojoty:
I merge this doc as it is now and we can continue discussion on details - I assume we agree in a general direction

following5:
> Yeah - our way of use composer is atypical - but in my opinion it fits best now - if we want to be sure that all nodes has same code - with same version of ext. libs we have two ways:
> * store ext.libs. code in repo - what exactly we do now
> * run composer update as part of update process and handle situation when it for example fails...
> 
> @following5 How it should work in your opinion?

Keep it like it is. :-) It is  bulletproof stable, and allows quick & easy switching between OC code versions on developers sites.

PR #1916, opened 2019-01-30
Merge mobile and desktop translations / add mobile to crowdin

following5:
This is a proposal for merging the mobile site translations to I18n translations, and thus adding them to Crowdin.

Merging this needs a special workflow, because it directly changes the translation files for languages pl, nl, ro and de:

1. Merge open Crowdin PR, if any.
2. Merge this.
3. Wait until en.php has been pushed to Crowdin.
4. Upload the pl, nl, ro and de translation files to Crowdin, without checking any options (i.e. add only the new translations).

A side effect of this change is that that https://m.opencaching.PL and http://m.opencache.UK will no longer be available in Romanian. But it can be added again later, when all translations have been switched from database to files.


following5:
The error page for unsupported language (e.g. https://opencaching.pl/?lang=xx) does not work with mobile page, because the desktop template system is not available there. It could be disabled for mobile views -- but: I propose to remove it completely and just use the fallback language.

Users will see that error page only if manually entering an unknown language in the URL. That will very rarely happen, because they need to know the "lang=xx" syntax. To know that, they need to click on one of the language flags. But then, they already know which languages are available or not, so why tell it again on a separate page?

Falling back to default language also makes it easier to link to OCPL from extenal pages, e.g. from OKAPI. See e.g. [this](https://github.com/opencaching/okapi/blob/master/okapi/core/Okapi.php#L1282), which will currently does not with ro, fr and de languages. E.g. c:geo authorization at OC RO opens an English login page instead of Romanian.

kojoty:
@following5 
We treat mobie version as a legacy project. As you can see this is not linked anywhere and we thought about drop it long time ago, but there are a few users of this so it is still online, but we do only absolutely necessary changes in that code.
We rather think about change main OCPL site to work better with mobile than continue that project.

My opinion is not to merge this PR.

following5:
@kojoty 
The translations of the mobile page have been broken by the recent I18n refactoring. E.g. see this in the English translation:
* http://m.opencaching.pl/viewcache.php?wp=OP0001

Type: Tradycyjna
Size: Normalna
Status: Gotowa do szukania

This change fixes that. What disadvantages would arise from meging it, which are greater than the benefit of fixing the translation?

kojoty:
@following5 please do what is the best in your opinion - I only wanted to share with you my opinion about mobile site. I didn't look nto code of this PR - if it fixes anything - please continue.

following5:
I should habe mentioned the issue/fix right away ...

It is hard to decide which is the best solution here, but my feeling is that a clean integration to I18n will make the least trouble with future refactoring.

following5:
Mobile page login has been broken for a long time, if it ever worked. It pretents to login, and then simply uses the main page session. Whatever user name and password is entered.

mobile

PR #1923, opened 2019-01-31
replaced util.sec/dbupdate by UpdateController

following5:
This new Update Controller does all updating that is necessary after a code deployment, including OKAPI update.

### DIFF
--- a/Controllers/UpdateController.php
+++ b/Controllers/UpdateController.php
@@ -1,0 +1,81 @@
+<?php
+
+namespace Controllers;
+
+use Utils\Database\DbUpdates;
+use Utils\Lock\Lock;
+use okapi\Facade;
+
+/**
+ * This class runs all updates that are necessary after a code deployment.
+ */
+
+class UpdateController extends BaseController
+{
+    public function __construct()
+    {
+        parent::__construct();
+    }
+
+    public function isCallableFromRouter($actionName)
+    {
+        // It MUST be safe to run update any time. (Concurrent runs are prevented
+        // by locks). Therefore it is also safe to publish this page via router.
+        // Note that the OKAPI update view has always been public.
+
+        // However, it may not be safe to run only parts of update. Therefore only
+        // the full update is runnable by router:
+
+        return ($actionName == 'index');
+    }
+
+    /**
+     * Run all updates including OKAPI
+     */
+    public function index()
+    {
+        ob_start();
+
+        // OC database update
+        echo "Run OC database updates\n";
+        echo self::runOcDatabaseUpdate();
+        echo "\n";
+
+        // OKAPI update
+        Facade::database_update();
+
+        $messages = ob_get_clean();
+        $this->view->showPlainText($messages);
+    }
+
+    /**
+     * Run all OC database updates that are ready to run
+     *
+     * @return string
+     *    multiline English plain text, diagnostic notices, should be displayed
+     *    to the operator / developer if the update was run manually.
+     */
+    public static function runOcDatabaseUpdate()
+    {
+        $lockHandle = Lock::tryLock('DbUpdate', Lock::EXCLUSIVE | Lock::NONBLOCKING);
+        if (!$lockHandle) {
+            throw new Exception('Database update is already running, or problem with lock file.');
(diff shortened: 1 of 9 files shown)


deg-pl:
@following5 I am sorry, I have a lot of work for the next 2-3 months and I am not able to deal with the development of the OC code at this time. Your code looks good, but please let @kojoty see it.

Please change http to https in the examples.

Controllers/UpdateController.php
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * This class runs all updates that are necessary after a code deployment.
+ */
+
+class UpdateController extends BaseController
+{
+    public function __construct()
+    {
+        parent::__construct();
```
kojoty: I assume we need some privileges check here - am I right?
This controller should be accessible only for users with sysAdmin roles probably...

following5: See this:

```
    public function isCallableFromRouter($actionName)
    {
        // It MUST be safe to run update any time. (Concurrent runs are prevented
        // by locks). Therefore it is also safe to publish this page via router.
        // Note that the OKAPI update view has always been public.

        // However, it may not be safe to run only parts of update. Therefore only
        // the full update is runnable by router:

        return ($actionName == 'index');
    }
```

http://opencaching.pl/okapi/update is public for six years now. :-)

kojoty: if it stay as is everyone (anonymously) can call UpdateController->index() and get the results 
do you think it is safe? if yes - I'm ok with it.

following5: Mmmh. Could there be an update that outputs security-relevant diagnostic messages? It's very unlikely, but possible. So better protect this view.

following5:
The security check is tricky. It either needs a special, non-default Apache configuration, which includes the caller's host name in $_SERVER[] and has performance penalty for DNS resolution. Or we must compare IP addresses, which may [fail](https://stackoverflow.com/questions/12435582/php-serverremote-addr-shows-ipv6) if the server has multiple addresses (like IPv4 and v6).

Also, making it public is useful for remote diagnostics, especially if there will be run-always updates.

Let's make it public, and require developers to not "echo" sensitive information in their DB updates.

kojoty:
@following5 I don't understand - what kind of checking you would like to do here?
How do you plan to use this method?

I assume that calling `UpdateContr. -> index()`  starts db update process - right?
I assume that it can be started:
1. from cron or other script on server as a CLI
2. from cron or other script on server by web call (wget or something)
3. by anyone in Internet by web

- options 1. is OK - no comments
- option 2. - I think it should be prohibited at all - do we need this way?
- option 3. - I think it is OK but user should be authorized and has proper privileged

following5:
> I assume that calling UpdateContr. -> index() starts db update process - right?

Right. And that view cannot be protected by OC user authorization, because when run from CLI via wget, there is no user login. It must be runnable without login.

So how to protect that view?

following5:
> to run from CLI I think we can implement the way of doing this... some kind of script which can include Controller and call its method...

[Here](https://github.com/opencaching/opencaching-pl/pull/1777#issuecomment-447212655) you proposed the opposite for Cron Controller: Remove the script and run it via wget etc. "of course it needs small refactor to allowing call by local only" - so what would that small refactor be?

Also please see the title of this PR and https://github.com/opencaching/opencaching-pl/pull/1892#issuecomment-459392551. Running from CLI means that it must be run as webserver user, which is unhandy for manual runs by developer or sysadmin.

following5:
> please don't spend too much time on it - this security case is really side issue.

A change between util.sec script and wget-call would need another update of all deployment scripts at all OC sites. Worst case that can mean to temporarily cut off a site from code deployment, because sysadmin is unavailable - see current OC RO problem.

So we better decide *now* what is the best solution and go for that.

following5:
Ok, let's go for the CLI script and stay with that. I will name it /bin/update.php.

following5:
ROTFL  :-)

https://opencaching.pl/post-commit.php 

That stuff IS public, no matter if we run it via CLI or wget.

Issue #1881, opened 2019-01-23
Emoticons from tinyMCE depends on tinyMCE path


kojoty:
tinyMCE inserts links to the emoticons
like:
```
<img src="lib/tinymce4/plugins/emoticons/img/smiley-laughing.gif" border="0" alt="�miech" title="�miech" />
```
but now these emoticons are present under path:
``` 
assets/js/libs/tinyMCE/4.9.2/plugins/emoticons/img/
```

if path to tinyMCE is changed all these path needs to be updated across all DB columns which can contains emoticons which are:
- geocache description 
- geocache log
- geopath description
- geopath logs
- ...?
We need simple script to run every time tinyMCE is updated which fix the paths or other solution for this.

following5:
Cache listings and logs can be stored offline, especially in data that was downloaded as GPX; logs may also copied to another geocaching website. Emoticons will also break in this replicated data. Therefore I would try to keep the icon path unchanged (TinyMCE version numbers may be omitted!). Many TinyMCE users will have the same problem, so maybe there is a setting which allows to store the emoticons at a certain path?

If it is unavoidable to change the emoticons path, then there are still some options to solve this:

* Do not completely remove the old TinyMCE installations, but keep the emoticons files.
* Create symbolic links for that. May be automatically done by an always-run "DB update".
* Bend the old emoticon URLs via Apache rewrite
* ... ?

deg-pl:
We have emoticons disabled in TinyMCE from over one year. Maybe we can replace in DB paths to emoticon images into UTF-8 emoticons?

kojoty:
> We have emoticons disabled in TinyMCE from over one year. Maybe we can replace in DB paths to emoticon images into UTF-8 emoticons?

good idea 
table of unicode emoticons: https://apps.timwhitlock.info/emoji/tables/unicode

following5:
GSAK will display ? instead of the emoticons.

following5:
Couldn't this be fixed by letting SimpleRouter redirect the old icons path? Would avoid editing cache and log contents *and* stay compatible with externel copies of the content.

kojoty:
Fixed by htaccess redirect.

PR #1924, opened 2019-01-31
Fix emoticons + start of folders structure cleanup

kojoty:
- after all I decided to use htaccess redirect way to fix the issue #1881
- there is new `/public` dir and first files inside
- index.php is moved to `/public` - necessary changes are in .htaccess files


.htaccess
```diff
@@ -1,33 +1,31 @@
...
+# redirect requests like /OP123F to viewcache page 
+RewriteRule ^(O[A-Z][0-9A-Z][0-9A-Z][0-9A-Z][0-9A-Z])$                              viewcache.php?wp=$1 [R=301,L,NC]
 
-# redirect requests like OP123F to viewcache page 
-RewriteRule ^(O[A-Z][0-9A-Z][0-9A-Z][0-9A-Z][0-9A-Z])$ viewcache.php?wp=$1 [R=301,L,NC]
-
-# block (403) requests to /mobile/*
-RedirectMatch 403 ^/mobile/$
+# block requests to /mobile/* - it works on another domain
+RewriteRule ^mobile/                                                                "-" [F]
```
following5: Error on my develsite when accessing local.m.opencaching.pl:

Forbidden

You don't have permission to access / on this server.

kojoty: this htaccess should not be used because for m.opencaching.* documentRoot should be in /srv/ocpl/mobile/ dir... does apache looks for htaccess in dir above documentRoot as well? 

I will look at this...

kojoty: OK, now I understand better how .htaccess works :)
This PR need fix - I'll prepare it tomorrow.

kojoty: still not resolved - I'm going to come back to this asap...

PR #1935, opened 2019-02-02
Refactoring of emails settings

kojoty:
This is simple refactor - I made little cleanup around emails addresses.

@harrieklomp @andrixnet 
please check if new configs for your nodes are OK

@deg-pl 
I need config for OCUK - can you prepare such file or send me current setting.inc.php from OCUK?

### DIFF
--- a/okapi_settings.php
+++ b/okapi_settings.php
@@ -3,7 +3,7 @@
 namespace okapi;
 
 use lib\Objects\OcConfig\OcConfig;
-use Utils\I18n\I18n;
+use Utils\Cache\OcMemCache;
 
 function get_okapi_settings()
 {
@@ -28,7 +28,6 @@ function get_okapi_settings()
     # OKAPI defines only *one* global variable, named 'rootpath'.
     # You may access it to get a proper path to your own settings file.
 
-    require(__DIR__.'/lib/settingsGlue.inc.php');  # (into the *local* scope)
 
     # save OKAPI classloaders
     $savedAutoloaders = spl_autoload_functions();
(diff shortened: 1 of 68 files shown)


harrieklomp:
Done for OCNL

deg-pl:
@kojoty - done

okapi_settings.php
```diff
@@ -37,26 +47,41 @@ function get_okapi_settings()
...
+
+        'FROM_FIELD' => OcConfig::getEmailAddrNoReply(),
         'DEBUG' => $debug_page,
         'DB_SERVER' => $dbserver,
         'DB_NAME' => $dbname,
         'DB_USERNAME' => $dbusername,
         'DB_PASSWORD' => $dbpasswd,
-        'SITELANG' => 'en', //TODO: how to read it from I18n class?
-        'SITELANGS' => array_map('strtolower', $config['defaultLanguageList']),
+        'SITELANG' => I18n::getDefaultLang(),
```
kojoty: @following5 Please take a look at this - this is my proposal of the solution for access to OCPL classes (OcConfig) to read new settings - what do you think?

I call OCPL autoloader and next restore OKAPI settings...

following5: I would rather avoid to call I18n and OcConfig code from OKAPI for performance reasons. OcConfig creates ApplicationContainer, ApplicationContainer creates database connection ... and whatever else may happen there.

Just working on another proposal which avoids that and is very extensible with null performance penalty.

following5: But I am fine with your above proposal as temporary workaround. Can remove the language workaround then from OKAPI code.

My proposal needs the UpdateController, whch - I guess - is not installed yet in OC PL and UK deployment scripts?

following5: ... temporary solution, not workaround. It will do it, just can be optimized in performance. :)

following5: @kojoty
OKAPI will need translations of country names etc. Currently quickly to fetch from `countries` table, but soon will require loading bulky OCPL translation files!

kojoty:
OK, so now here we have:
- simple OCConfig class (without any external depredations)
- simple okapi_settings - just calls to OCConfig

@following5 
can we merge this PR now?

following5:
I get this error when testing OKAPI:

`Fatal error: okapi\get_okapi_settings(): Failed opening required '/srv/ocpl/lib/settingsGlue.inc.php' (include_path='.:/usr/share/php') in /srv/ocpl/okapi_settings.php on line 30`

following5:
There was some very stange problem with Git checkout on my host system. Ok after reboot.

Thumbs-up for the okapi_settings change.

PR #1938, opened 2019-02-03
fix geocache attribute title lang issue

kojoty:


kojoty:
this is very simple fix so I merge during the weekend

PR #1941, opened 2019-02-04
new interface for passing settings to OKAPI

following5:
Creates `/var/okapi_settings.inc.php` with each code deployment, which is included to `/okapi_settings.inc.php`.

* maximum OKAPI performance, no matter how expensive it may be to provide the OCPL data;
* minimum code dependency OCPL => OKAPI.

The generated file now looks like this:

```
<?php

return array (
  'OC_NODE_ID' => 2,
  'SITELANG' => 'pl',
  'VAR_DIR' => '/var/www/html/ocpl/var/',
  'OCPL_ENABLE_GEOCACHE_ACCESS_LOGS' => false,
);
```

These are the OKAPI settings that are already available via OcConfig and I18n. More will be refactored to here from `/okapi_settings.php` (email addresses next).

### DIFF
--- a/Controllers/UpdateController.php
+++ b/Controllers/UpdateController.php
@@ -4,6 +4,7 @@ namespace Controllers;
 
 use Utils\Database\DbUpdates;
 use Utils\Lock\Lock;
+use Utils\I18n\I18n;
 use okapi\Facade;
 
 /**
@@ -43,6 +44,7 @@ class UpdateController extends BaseController
         echo "\n";
 
         // OKAPI update
+        $this->updateOkapiSettings();
         Facade::database_update();
 
         $messages = ob_get_clean();
@@ -78,5 +80,25 @@ class UpdateController extends BaseController
         }
 
         return $messages;
+    }
+
+    /**
+     * Create settings file for OKAPI
+     */
+    private function updateOkapiSettings()
+    {
+        // See explanation in /okapi_settings.php.
+
+        $settings = [
+            'OC_NODE_ID' => $this->ocConfig->getOcNodeId(),
+            'SITELANG' => I18n::getDefaultLang(),
+            'VAR_DIR' => $this->ocConfig->getDynamicFilesPath(),
+            //'OCPL_ENABLE_GEOCACHE_ACCESS_LOGS' => $this->ocConfig->isCacheAccesLogEnabled(),
+            'OCPL_ENABLE_GEOCACHE_ACCESS_LOGS' => false,
+        ];
+
+        $code = "<?php\n\nreturn " . var_export($settings, true) . ";\n";
+
+        file_put_contents(__DIR__.'/../var/okapi_settings.inc.php', $code);
     }
 }
--- a/okapi_settings.php
+++ b/okapi_settings.php
@@ -27,36 +27,36 @@ function get_okapi_settings()
 
     require(__DIR__.'/lib/settingsGlue.inc.php');  # (into the *local* scope)
 
-    return array(
-        # These first section of settings is OKAPI-specific, OCPL's
-        # settings.inc.php file does not provide them. For more
-        # OKAPI-specific settings, see okapi/settings.php file.
+    return [
+        # OKAPI-specific settings which are not taken from OCPL settings.
+        # For more OKAPI-specific settings, see okapi/Settings.php file.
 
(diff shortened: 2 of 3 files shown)


following5:
@deg-pl 
see https://github.com/opencaching/opencaching-pl/pull/1923/commits/21de6fd4aa412c9b6fe7acbb943eec93eb069485#diff-4

This will enable the new database update system.

deg-pl:
@following5 should be OK now.

following5:
Yep, thanks.

kojoty:
@following5 
yeah, this seems to be a good idea.

...but if we are afraid about performance of generating okapi-settings on every request by OCPL classes I think we can extend the idea a little:
 - if OKAPI settings is auto-generated we can as well store in in APC cache instead of in file
- every OCPL node has APC configured and worked
- then it will be fast (maybe even faster that reading a file from disk) and mych more independent
- APC has build-in timeout so settings can be refreshed - let's say every 5 min.

what do you think?

following5:
@kojoty 
The file will be cached in memory because it is frequently requested. I assume it will be cached in compiled form in the Apache PHP module. So there would not be a performance advantage by APC.

Also, APC maybe does not run stable at OC UK: https://github.com/opencaching/opencaching-pl/pull/1811#issuecomment-451642232

And APC cache is cleared with every system reboot. This would need additional setup at OC sites and at devel VMs for restoring the file after reboot.

kojoty:
> The file will be cached in memory because it is frequently requested. I assume it will be cached in compiled form in the Apache PHP module. So there would not be a performance advantage by APC.

Yes, that's can be right.

> Also, APC maybe does not run stable at OC UK: [#1811 (comment)](https://github.com/opencaching/opencaching-pl/pull/1811#issuecomment-451642232)

Nope, APC is (in my opinion) production standard in current PHP - everything can be unstable - but generally we used APC in many places right now and it works OK.

> And APC cache is cleared with every system reboot. This would need additional setup at OC sites and at devel VMs for restoring the file after reboot.

Nope - It can be done like that;

okapi_settings.php now just provides a function which returns static array
I postulate the change:
- okapi_settings will try to load this static array from APC
- if there is no such data in APC cache - this static array could be created and saved in cache
- next request to okapi_settings use data from cache...
so only rare request will need to load all OCPL stuff

following5:
I have done some performance testing. Loading of OKAPI settings currently takes 0,03 ms on my develsite, which tanslates to < 0,01 ms on a production site. Compared to the average ~25 ms runtime of OKAPI methods at OC PL this is neglectable. Calling optimized OcConfig methods won't make much a difference.

So I suggest to proceed with your first proposal and postpone any optimizations of okapi_settings.

following5:
@kojoty 
https://github.com/opencaching/opencaching-pl/pull/1935/commits/7a98db9a96afffc2bee67caaeeb16cf34b99c982#diff-7c237efda511397e1a8e86701e43f5edR38

The cached settings must be discarded, if OCPL settings have changed. OKAPI may malfunction if it uses outdated settings from APC cache.

kojoty:
@following5 I'm aware of it - this is just a proof of concept... I'm not going to merge it without cooperation with you :)

following5:
Btw, it is not possible to use any OCPL classes in okapi_settings.inc.php, as long as the OCPL autoloader is configured via okapi_settings.inc.php:

```
'EXTERNAL_AUTOLOADER' => __DIR__.'/lib/ClassPathDictionary.php',
```

The solution that I proposed above would work around that problem. Otherwise, the OCPL autoloader path must be defined directly in OKAPI code.

kojoty:
@following5 Let's conclude: what is in your opinion the best solution for okapi_settings - is this PR is ready to merge then?

If yes let's merge it and then I rebase and merge my changes - of course I revert changes around okpai_settings in PR #1935

following5:
Umm... merging this would break OKAPI settings at OC RO, and according to my performance tests, it does not improve anything.

I don't know what's the best way to proceed, and I don't know how to fix the APC-cache problem mentioned above. Will move the autoloader path into OKAPI code, so that OcConfig and I18n may be accessed from okapi_settings.inc.php. Then please go on as you prefer.

following5:
> Will move the autoloader path into OKAPI code

done.

PR #1956, opened 2019-02-06
Whitespace cleanup - normalize line breaks to LF

following5:


following5:
This will clash with #1965.

kojoty:
@following5 Don't worry - I'll try to rebase - or if you made it with script we can try to do it after commit that huge PR.

PR #1957, opened 2019-02-06
add cache link to adoption message; closes #1950

### LINKED ISSUE #1950: Adoption successful banner should contain link to cache
When adopting a cache, the receiving user, upon accepting the cache, will see a green banner saying that the cache was successfully adopted. The cache name is shown, but it is a plain text string.

It should be also a link to the cache page.

following5:


lib/Objects/GeoCache/GeoCache.php
```diff
@@ -1706,4 +1708,19 @@ public static function nanoIsInUse()
             self::SIZE_NANO
         ) != 0;
     }
+
+    public function getAbsLink()
+    {
+        return SimpleRouter::getAbsLink('ViewCache') . '?wp=' . $this->getWaypointId();
```
kojoty: there already is similar method:  getCacheUrl()

Utils/View/View.php
```diff
@@ -61,6 +61,11 @@ public function setVar($varName, $varValue)
         $this->$varName = $varValue;
     }
 
+    public static function makeLink($uri, $plainText)
+    {
+        return '<a href="'.$uri.'">'.htmlspecialchars($plainText).'</a>';
```
kojoty: I'm softly against such solutions.
- It mix the code with the view. 
- It seems to be generic, but in fact you often need to add some unique id of class to HTML anchor and then there needs to be lots of params 
- if the HTML is more complicated or we need to uniform calling something (for example we want one place in code where some JS lib is included) there are "chunks":https://github.com/opencaching/opencaching-pl/tree/master/tpl/stdstyle/chunks

Controllers/CacheAdoptionController.php
```diff
@@ -178,7 +178,7 @@ private function actionAccept(GeoCache $cacheObj)
         $this->db->commit();
 
         $message = tr('adopt_15');
-        $message = str_replace('{cacheName}', $cacheObj->getCacheName(), $message);
+        $message = str_replace('{cacheName}', $cacheObj->getLinkedName(), $message);
```
kojoty: now we have also printf-like style of inserting vars to translations so it can be refactored to:
```
$message = tr('adopt_15', $cacheObj->getLinkedName());
```
and the translation phrase can be like:
`bla-bla-bla %s yeah`

following5: Yep. This can be done for some hundert of those variables, which is a task of its own.

Utils/View/View.php
following5: Actually this is View code. ;-) But agree that it's not optimal. Will leave the issue to someone else who is more experienced with OCPL code architecture.

lib/Objects/GeoCache/GeoCache.php
deg-pl: Yes. Better way is to extend current method - sth like this:
getCacheUrl(bool $absolute = false)

PR #1965, opened 2019-02-09
Big refactor to new folders structure 

kojoty:
I think we need to do it one day so maybe today :)
I'm not sure if everything works - but I try to test as much as I can - the rest we fix "in production"

t.b.c.


kojoty:
BTW. this is not ready to commit.. yet

kojoty:
@following5 @deg-pl @andrixnet @harrieklomp 
I'm going to merge this PR today evening. This is huge number of changes in paths so I'm not 100% sure if everything works after merge - I will try to test as much as possible and fix the rest after merge.

kojoty:
@harrieklomp @andrixnet 
After merge this PR it is necessary to drop all APC cache entries - I clean cache for all nodes I have access to APC (NL/UK/PL) but I have no access to RO service...

PR #1979, opened 2019-02-26
adjust 100_init db update for old mysql/mariadb

following5:


following5:
Fix for initializing DB update on OC RO.

PR #1993, opened 2019-03-08
fasts fix for late-night loggers

following5:


kojoty:
@following5 thanks for fixing my bugs!

PR #2010, opened 2019-04-05
Make map working for OCUS

harrieklomp:
This is a simple map config to make maps working on OCUS (DEVEL)

### DIFF
--- a/config/map.us.php
+++ b/config/map.us.php
@@ -1,0 +1,97 @@
+<?php
+
+/**
+ * $map['jsConfig'] constains JS code with configuration for openlayers
+ * Be sure that all changes here are well tested.
+ */
+$map['jsConfig'] = "
+{
+    OSM: new ol.layer.Tile ({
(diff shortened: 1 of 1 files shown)


config/map.us.php
```diff
@@ -0,0 +1,97 @@
...
+                'World_Street_Map/MapServer/tile/{z}/{y}/{x}'
+        })
+    }),
+}
+";
+
+/**
+ * Bing map key is set in node-local config file
+ */
+$map['keys']['BingMap'] = 'NEEDS-TO-BE-SET-IN-LOCAL-CONFIG-FILE';
```
kojoty: if there is no plan to use bing maps (I can't see it in config) there is no need to override 
`$map['keys']['BingMap']`
and 
`$map['keyInjectionCallback']` as well

harrieklomp: I do not know if OCUS is planning to use Bing maps. If yes, they should generate a key for this.
I will contact OCUS :-)

PR #2047, opened 2019-05-18
latest logs fixes

kojoty:


kojoty:
exceptional commit in weekend - fix of the very expensive query + smaller fixes for only one view

harrieklomp:
It seems that OCNL and DEVEL sites NL and US have a problem with this change.

kojoty:
should be fixed now

PR #2054, opened 2019-05-22
cache_additions DB indexes cleanup

kojoty:
this is my first, simple DB update
cache additions should has primary index instead of unique

@following5 I have never used your new engine for DB update before - this is great - chapeau bas


kojoty:
This PR is waiting for review and verification - according to the following idea every DB change PR should be treated especially

following5:
The only thing which needs to be checked are the SQL statements. Everything else is auto-generated, you can't get it wrong.

kojoty:
> you can't get it wrong.

I'm ready to bet that's I'm able to do it wrong :) 
(but that's right it needs acting in a bad will)

PR #2084, opened 2019-07-18
Modified cache_size table

harrieklomp:
In order to make it possible for nodes to add nano cache size on there sites here a update for table cache_size. This also includes all available languages.
Please check it and run this sql alter on OCPL and OCUK

### DIFF
--- a/sqlAlters/cache_size.sql
+++ b/sqlAlters/cache_size.sql
@@ -1,0 +1,40 @@
+-- 2019-07-17
+-- @author: harrieklomp
+--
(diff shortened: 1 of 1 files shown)


andrixnet:
Considering https://github.com/opencaching/opencaching-pl/issues/2085 this change has all the chances of not having the desired effect...

andrixnet:
Changing this table has no effect according to tests. 
Better clean up the code of all remaining references to this table: https://github.com/opencaching/opencaching-pl/issues/2085 
Thus it's better closed.

Issue #2076, opened 2019-07-04
GeoPath FINAL cache type


andrixnet:
There is a currenttly unused (dropped) cache type : type 8 (podcache). 
This has already been partially redefined as GeoPath FINAL (todo). 
Currently it is/seems unused and by default not allowed, also it still features the old podcache graphics. 

The purpose of this issue is to track the complete  implementation of GeoPath FINAL cache type.

### Stage 1
- [ ] cache type definition - in code
- [ ] cache type definition - in database (as it still exists and is used)
- [ ] remove old podcache data
- [ ] update database contents, all caches with ID 8 => ID 1 (other) + MP3 attribute
- [ ] create new graphics for this type
At this point the cache type is properly defined. Type remains disabled by default.

### Stage 2
- [ ] define the workings of this cache type
- [ ] implement the corresponding code
- [ ] enable the type on production servers (where geopath module is active)

andrixnet:
Assigning myself for stage 1 as part of https://github.com/opencaching/opencaching-pl/issues/2024 and https://github.com/opencaching/opencaching-pl/issues/860

kojoty:
BTW: I think that there is no reason for geopath final - in my opinion this type is generally useless.

andrixnet:
If such a type is deemed useless, no problem. I'll make the changes so that it is not included. 
Please close this as WON'T FIX if this type should not be implemented.

kojoty:
but we need to be sure that this is removed from the rest of the code... - please let it be open for now.

deg-pl:
I think we should to remove "geopath final" from OC code.

andrixnet:
Removal in progress

andrixnet:
Removed at https://github.com/andrixnet/opencaching-pl/tree/cache_type

Please see https://github.com/opencaching/opencaching-pl/issues/860#issuecomment-508733416 for remaining details.

andrixnet:
Database contents update is designed to work in conjunction with Refactoring attributes PR (uses new attrib ID).

Issue #2074, opened 2019-06-29
ErrorException: Undefined offset: 12


andrixnet:
After solving many other problems, I keep getting these at OCNA server.
```
ErrorException: Undefined offset: 12



#0 /path/to/oc_us/oc/lib/myn.inc.php(75): src\Utils\Debug\ErrorHandler::handleError(8, 'Undefined offse...', '/path/to/oc_u...', 75, Array)

#1 /path/to/oc_us/oc/newcaches.php(120): myninc::checkCacheStatusByUser(Array, NULL)

#2 {main}
```
I am at a loss PHP-wise

kojoty:
It seems that there is not cache type icon ==12 in /lib/cache.php

I can fix it but after the weekend. Now I'm again on the trip...

andrixnet:
I'll tend to it (at least the graphics) early this week.
It is seems directly related to https://github.com/opencaching/opencaching-pl/issues/860 and https://github.com/opencaching/opencaching-pl/issues/2024

Issue #2024, opened 2019-04-25
CACHE_TYPE_ID changes


andrixnet:
Migrating OCNA also means importing several new/different cache types ( part of #860 ), some of which have overlapping IDs with opencaching-pl code.

As detailed in the documentation here: https://wiki.opencaching.eu/index.php?title=Cache_types
there are several cache types for which the ID must be changed to ensure uniqueness, which is critical for consistent code and deployment across sites.

The following changes are needed: 
- [ ] Moving cache ID = 8 -> new ID = 20 (overlapping with OCDE)
- [ ] Geopath final cache ID = 9 -> new ID = 21 (overlapping with OCDE and OCNA)
- [ ] MP3 cache ID = 9 -> new ID = 22 (overlapping with OCDE and OCNA)
- [ ] Owncache ID = 10 -> new ID = 23 (overlapping with OCDE and OCNA)
- [ ] Guestbook cache ID = 10 -> new ID = 24 (overlapping with OCDE and OCNA)
- [ ] BIT cache ID = 12 -> new ID = 25 (OCNA only)
- [ ] Benchmark ID = n/a -> new ID = 26 (OCNA only)

While some overlaps are noted with OCDE only, I believe it is best practice to ensure uniqueness overall. 
--------------------
Things needed to be done: 
- [ ] code changes where these IDs are in the code. At least cache types, possibly elsewhere
- [ ] database contents - existing data must be updated. `caches` table first, then all other tables that reference cache_type_id
- [ ] deployment on all nodes, code and database contents change.

There are a few caches unique to OCNA, but the same queries will also update the contents of their existing caches.

I am studying the database layout and contents, and will attempt to produce SQL queries to make the changes. 
I am asking for help for code changes, especially since while searching for places where changes are needed I might miss some.

Once the queries are created, I am also asking for help in deploying the database changes using the new method. This is a small but important change and it changes contents of caches. I desire to proceed with utmost care and being my first time generating a DB change with the new method I am reluctant to do it on this scale.

Thank you.

andrixnet:
Assigning myself as first stage to develop the SQL queries.

andrixnet:
Further work in progress: https://github.com/andrixnet/opencaching-pl/commit/70079346709057e6069866e8a703d6b7b2820036

Please also see questions here: https://github.com/opencaching/opencaching-pl/issues/860#issuecomment-508586654

andrixnet:
[...]

Issue #2013, opened 2019-04-08
TODO list for OCUS


harrieklomp:
To prevent there are many issues about immigrating Polish code to OCUS i suggest to all to add here what has to be done and to be solved to make it to a success. If you like to add or solve something, just add it and add your name to it :-)

- [ ] Make top menu including some site specific articles pages. Can be done with specific links in menu. (Harrie)
- [ ] Add welcome text. This should be done with OCUS admins. (Harrie)
- [ ] Make Geopath work with settings desired by OCUS admins. (Harrie)
- [ ] Add mascot to system.
- [ ] Add specific US items to left menu. (Harrie)
- [ ] setup of cron jobs (Andrei)
- [ ] French and Spanish translations. We need Spanish translator (Harrie)
etc, etc

andrixnet:
Almost ready for production release of new site, albeit with some known issues still to work on.

kojoty:
@andrixnet where is the complete list of issue (is this list above complete and updated?)
Thanks @andrixnet for all your work in this mater :)

andrixnet:
The list here contains many management and administrative tasks related to the migration process, in addition to the code development tasks. 
Most of the actual development issues are listed here: https://github.com/opencaching/opencaching-pl/projects/5 and here https://github.com/opencaching/opencaching-pl/milestone/1 and here https://github.com/opencaching/opencaching-pl/labels/OCNA%20migration
(I think they partly overlap, partly complete each other.)
-----------------------------------------
Main site is up and operational at http://www.opencaching.us (Aruba small VPS)
Still TODO: set up SSL certificates and configuration.
Email subsystem up and running.
Secondary sites up and running.
-----------------------------------------
Several configuration and user roles/rights still need to be configured. 
Some small details I managed already to sovlew directly. 
I am in constant contact with our american friends.

andrixnet:
@kojoty when you have the time, please first take a look at https://github.com/opencaching/opencaching-pl/pull/2069 

This paves the way for the first time unified attributes between OCPL sites (and first step to move attribute definitions to code). Also the proper updates have been prepared for OKAPI and a corresponding pull request is pending there. 

Attributes are now well documented here https://wiki.opencaching.eu/index.php?title=Cache_attributes
I am also in contact with c:geo developers to support the complete OC attribute set.

Next step after this PR will be to add/remove some attribs at NL, RO, UK, US according to individual requests from each OC site team.

[...]

PR #2087, opened 2019-07-23
Cache type changes (OCNA migration)

andrixnet:
This is my work related to cache type changes according to: 
[CACHE_TYPE_ID changes](https://github.com/opencaching/opencaching-pl/issues/2024), [TODO list for OCUS](https://github.com/opencaching/opencaching-pl/issues/2013), [GeoPath FINAL cache type](https://github.com/opencaching/opencaching-pl/issues/2076), [ErrorException: Undefined offset: 12](https://github.com/opencaching/opencaching-pl/issues/2074).

- this changes the ID numbers for several cache types (exotic types, seldom used) to ensure no overlap occurs between OC branches, according to documentation [here](https://wiki.opencaching.eu/index.php?title=Cache_types)
- implements in code and database dependencies the new cache types requested by OCNA
- adds graphical resources for the new cache types.
- removes Geopath FINAL type

Notes:
- OKAPI has corresponding PR with updates regarding this. 
- c:geo has been notified here https://github.com/cgeo/cgeo/issues/7714
- GeoKrety has been notified [here ](https://github.com/geokrety/geokrety-website/issues/427)(due to OCXML exposing internal IDs)

PR consists of code changes and database update.
Database backup is recommended before running the SQL file.
SQL file is for manual run by admin and is valid for all member sites. 
This code can be seen in action at this devsite: http://ocro.dev.andrix.eu/ until merged.

Between code merge and DB update a few cache types may not work or be displayed incorrectly due to incomplete database definition.


harrieklomp:
Just wondering. Is there any progress in this PR?

Amberel:
Can I call you later?

Rgds, Andy

harrieklomp:
> Can I call you later? Rgds, Andy
Yes, sure :-)

andrixnet:
@kojoty @deg-pl @following5  can this be merged? 
In conjunction with corresponding OKAPI PR: https://github.com/opencaching/okapi/pull/595 ?

Let's schedule the merge this week.

kojoty:
OK, I will try to manage to review it today...

kojoty:
@andrixnet you made great work - my respect :)

I have a few questions:

#1: I see a few new types from OCNA:
>   - Guestbook cache ID = 10 -> new ID = 13 (overlapping with OCDE and old OCNA)
>   - BIT cache ID = 12 -> new ID = 12 (OCNA only)
>   - Benchmark ID = n/a -> new ID = 14 (OCNA only)
>   - Challenge ID = n/a -> new ID = 15 (OCNA only)

Please explain for me what these types means - I have a dirty idea: maybe they doesn't need these types as much and we can go other way - just refactor OCNA caches in DB to "our" well defined types...
(sorry that I ask at the end of the process - I know I should have asked about it long time ago...)

#2: have you tested this?
I mean what will happen when we merge this with OCPL server - are you sure that if we just add these new types as forbidden there will not be any problem - for now I afraid of this...
I will also try to test it on my dev...

kojoty:
OK, I see the problem:
- forbidden types in config is only to prevent creating new caches of such types - but it not hide not used caches... 

At least at map I see now:
image

andrixnet:
@kojoty these cache types are explained briefly here: https://wiki.opencaching.eu/index.php?title=Cache_types 
I am planning a unified documentation there, with each cache type described in more detail, like
https://wiki.opencaching.eu/index.php?title=Size:Micro or
https://wiki.opencaching.eu/index.php?title=Log:Attended
It is on the TODO list :-) 

Additional information (though not fully up to date) can be found here: 
https://wiki.opencaching.us/index.php/Cache_parameters#Cache_type

Also, you can see the ID overlaps in this older version of the wiki page:
https://wiki.opencaching.eu/index.php?title=Cache_types&oldid=680 (before numbering refactor).

You can see them in action on one of my dev sites here: http://ocro.dev.andrix.eu/
This devsite is running the code from this PR (including RO data with SQL updates) and a few test caches using the new types.
As far as I managed to test myself (and perhaps Harrie), seems ok.
Note: this devsite does not include the corresponding OKAPI PR.

kojoty:
@andrixnet, Your work on unification types and attributes etc.  and documentation of all of that is _really_ impressive - I appreciate it (sure!) - many thanks for that.


I reviewed the new cache types from OCUS - sorry again that I made it so late (that was something what I was not analyze deeper before - I feel bad with this because you made lot of works and at the end I start discussion on that) -  anyway - my thoughts:

- BIT Caches - is in fact a specific type of virtual cache (or mobile if it can be moved - I'm not sure how it works) - you need to find a sticker with the code - right? - in my opinion this should be just virtual (or mobile if can be moved)  + maybe some attribute - I don't understand why we want to recognize it as a separate type of cache

- Guestbook - same - it can be traditional (if gestbook is on coords) or multi (if there are a few stages + gestbook at the end) or quiz (if you need to solve puzzle to find guestbook) or other (if this is not on coords) etc... I think it can be attribute of the cache - not a type of cache... - in fact this is just a specific type of logbook... 

- Benchmark - same - if this is a point under coords it is traditional - we can add attribute with information, that theses coords are confirmed or something (I don't understand this type...)

- Challenge - same as above - challenge means what - I need to do something hard to get the cache like climb on the wall to grab the box or don't sleep 3 days and then meet the owner? It can be traditional or own or quiz or any other type of cache - challenge should be an attribute!

**My idea for now:**
- let's merge the PR with unification of attributes - I understand that there was a big mess between nodes (number overlapping etc.) (am I right?) and now we will have a clear list of that + the simple way to extend that list (another thing is that we should migrate it from the db to the code - but this is next step)

- let's discuss if we really want to introduce the new types - in my opinion these are not "logical" and it will make system more complicated and less clear (virtual is a virtual or bit-cache if this is a sticker with a code... ?!) 

- even if we want to support the new types from OCUS we need to implement configurations which allow to DISABLE not supported types (at least for list in search page + filters on the map) for the node (I guess that our COG even doesn't know about new types and I think that RT shouldn't just introduce something like that without discussion with them)

@deg-pl, @harrieklomp @andrixnet @mzylowski 
please take a look at this discussion - please share your thinking! 

PS. 
@andrixnet again - sorry that I understand what is going on here and react so late :)

andrixnet:
@kojoty regarding cache types: I will try to relay things as requested by OCNA team: 

- **BIT caches**: best documented here: https://wiki.opencaching.us/index.php/BIT_Caches 
Physical object, though not container with log. Sticker usually also features log password. Not mobile. 
Closest equivalent IMO would be Munzee.
Explicitely requested by OCNA and a pre-existing cache type that they had implemented in their old site.

- **Guestbook**: it is a new or existing guestbook for visitors to sign it, both geocachers and non-geocachers. https://wiki.opencaching.us/index.php/Guest_Book_Cache 
It has been requested by US people explicitely as type and a pre-existing cache type that they had implemented in their old site.. (they rejected the attribute)

- **Benchmark**: physical object as geodetic benchmark, this can be looked at as a special kind of virtual since it has no container with log. 
Explicitely requested by US people, since the concept seems to have a special meaning for them and they knew about the attribute and rejected such variant.

- **Challenge**: meeting a specific geocaching challenge like finding for example all caches in a given area, etc. The bounds for possible challenges are still being worked out by the US team. AFAIK the given coordinates have no explicit meaning to visit and there is no physical container with log (unless stated otherwise), logging a find requires one to provide proof of meeting the requirements. 
Up to now US used "Unknown" type for this purpose, but they requested it as type.

-----------------------

**OCNA also had these cache types:** 
- **USB dead drop**: replaced by tradi/multi/puzzle with appropriate attribute

- **Podcast**: replaced by tradi/multi/puzzle with appropriate attribute 
(note this was also an old type at opencaching-pl, AFAIK it's code was reused for some time for a Geopath bonus, then it was dropped)

- **Letterbox**: replaced by tradi/multi/puzzle with appropriate attribute 

I have discussed all these in depth with the US team. They requested the former be as types and agreed about the latter to be assimilated with attributes.


-----------------------

- Attributes was a HUGE mess. It has PR here and PR at OKAPI. I have also submitted these as future changes to c:geo, so after we merge, they should be able to integrate easily.

- Attribute cleanup: I did the entire renumbering in one step, however there are some requests from several sites (such as US, UK, NL, RO) to add/remove attributes. I am planning on doing this (including work on attribute graphics) as a later stage. 

- I totally agree and support disabling certain cache types both as new hides and as search option for sites that do not desire to support such cache types. As far as my code and coding knowledge goes, it only works up to not allowing as new hide, they still show up as search criteria. 
If we want to remove them from search criteria, another restriction should be implemented, but this is beyond my skills. 

That is: 
1. cache types allowed as new hide -> they show up in new cache page and edit cache page

2. cache types allowed to be listed -> they show up everywhere (in lists, as seach criteria, etc)
These can be more then those at 1 to allow the display of existing caches of types no longer hideable.

One specific detail would be to allow editing an existing cache (of type not hideable) with the ability to retain the type.

andrixnet:
AFAIK this https://github.com/opencaching/okapi/pull/595 should be fine at OKAPI, several points raised by following5 have been resolved.

andrixnet:
Please review this PR in the light of the code changes this past year. Problems with cache types that persist and significantly affect OCNA are solved.
Thank you.

andrixnet:
A short review (for other purposes) today revealed that there still are many instances where `cache_type` table is still actually used. See here: https://github.com/opencaching/opencaching-pl/issues/1394#issuecomment-826896681 
Please also take this into account while reviewing this PR.

andrixnet:
IMO, given the changes in acceptance, changes in the code and strategy, this is best closed. 
Work done for this PR needs to be reconsidered in a major way and results submitted separately.

PR #2111, opened 2019-09-29
geopath candidates refactor (final changes)

kojoty:
instead of links in emails now there are just a list of:
- cache-candidates to given geopath: **/GeoPath/candidatesList/<gp-id>**
- list of invitations for current user: **/GeoPath/myCandidates**


kojoty:
I realize that this is not ready - I needto update cron job + db update scripts...

PR #2154, opened 2020-03-07
cache edit refactoring: part I - incomplete template

kojoty:
I've started the cache edit refactor as there are a few issues around this:
- problem with support for new types of caches (OCUS)
- other minor issues (ask me for details)

Currently this is side work - nothing is changed from user perspective - old script is workin until we decide that the new one is ready.

Finally the same template will be use for newcache and editcache dialogs.
I don't want to introduce many new features now - rather rewrite this code (editcache/newcache). 
Then adding the new features will be much easier.

This commit contains new template but there are still many open tasks - especially there is needed new JS code etc. I'd like to commit this to split the change on smaller parts.

### DIFF
--- a/src/Models/GeoCache/GeoCacheRules.php
+++ b/src/Models/GeoCache/GeoCacheRules.php
@@ -1,0 +1,94 @@
+<?php
+namespace src\Models\GeoCache;
+
+/**
+ * This class contains rules of which decide about allowed actions like:
+ * - switching the geocache status
+ * - log adding depends on log type
+ * etc.
+ *
+ * https://docs.google.com/spreadsheets/d/1BpkpmusS98HQKmyJJpJhCQtKHOltdY1b7QE6f8poW8E/edit#gid=1336776289
+ */
+class GeoCacheRules extends GeoCacheCommons {
+
+    /**
+     * Return array of geocache statuses which can be the next status based on given current status
+     *
+     * @param int $currentStatus
+     * @param boolean $isOwner
+     * @param boolean $isOcTeamMember
+     * @param boolean $isUserUnderVerification
+     *
+     * @return array of all possible next statuses (if empty no edit is allowed)
+     */
+    public static function getAllowedNextStatus ($currentStatus, $isOwner, $isOcTeamMember=false,
+        $isUserUnderVerification=false)
+    {
+        if (!$isOwner && !$isOcTeamMember) {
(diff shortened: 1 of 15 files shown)


kojoty:
link to new template working version (no there are no css for now):
`***/cacheEdit/editor/<cache-id>`

andrixnet:
Might I add a few notes?
- date picker same as cache publish page
- save changes on ENTER in any <input type=text>

src/Models/GeoCache/GeoCacheRules.php
```diff
@@ -0,0 +1,94 @@
...
+     * Return array of geocache statuses which can be the next status based on given current status
+     *
+     * @param int $currentStatus
+     * @param boolean $isOwner
+     * @param boolean $isOcTeamMember
+     * @param boolean $isUserUnderVerification
+     *
+     * @return array of all possible next statuses (if empty no edit is allowed)
+     */
+    public static function getAllowedNextStatus ($currentStatus, $isOwner, $isOcTeamMember=false,
```
rapotek, line 24: I suggest `getAllowedNextStatuses` or `getAllowedNextStatusList`: this method returns an array of allowed statuses.

src/Models/GeoCache/GeoCacheRules.php
```diff
@@ -0,0 +1,94 @@
...
+                    $next[] = self::STATUS_UNAVAILABLE;
+                    $next[] = self::STATUS_ARCHIVED;
+                    $next[] = self::STATUS_NOTYETAVAILABLE;
+                    $next[] = self::STATUS_BLOCKED;
+                }
+                return $next;
+        }
+
+        // never should be here...`
+        return [];
```
rapotek, line 91: I suggest `return $next`: `$next` is either returned earlier or an empty array here.

src/Models/GeoCache/Mp3Attachment.php
```diff
@@ -0,0 +1,225 @@
...
+use src\Models\BaseObject;
+use src\Utils\Debug\Debug;
+use src\Models\User\User;
+
+/**
+ * Generic representation of picture atahed to log/cache/...
+ *
+ */
+
+class Mp3Attachment extends BaseObject
```
rapotek, line 13: Just a question: shouldn't it be rather a general file attachment model class instead of MP3 only? I assume, the processing would be the same regardless of an attachment type.

Issue #2164, opened 2020-05-14
Results of search are not displayed but query is called


mzylowski:
During searching a cache or user on our site sometimes table with results is not returned:
![image](https://user-images.githubusercontent.com/13780868/81905810-8fb01900-95c5-11ea-93cc-bda4d9d6cef9.png)

The number of results is visible so I believe results also are taken from DB but are not displayed.

It is also reproducible in another places like:

>  wejdźcie w profil dowolnego użytkownika, który zakłada skrzynki (na przykład w swój). Zjedźcie do sekcji "Statystyka liczbowa skrzynek założonych / Liczba wszystkich skrzynek użytkownika" i kliknijcie w przycisk [Pokaż]. 

In some cases issue is sporadic - sometimes works fine, sometimes not (for example for next search this should work). 
But for example for list of user caches (clicked from user profile) the reproduction is stable:
Example ->[click here](https://opencaching.pl/search.php?showresult=1&expert=0&output=HTML&sort=bycreated&ownerid=99784&cachetype=1111101111&searchbyowner=&f_inactive=0&f_ignored=0&f_userfound=0&f_userowner=0)

mzylowski:
Probably also this is a reason of another issue related with Badges:
![image](https://user-images.githubusercontent.com/13780868/81907359-e7e81a80-95c7-11ea-99f1-d0954bdeee52.png)

rapotek:
Looks like it's a problem with GCT.js. It is written in a synchronous way, i.e. it assumes that after `google.load('visualization', ...)` everything is available at once, while it is not always true.
The correct way should be a call of `google.setOnLoadCallback(the_whole_GCT_function)`, but with code organization in search templates, refactoring may be not trivial.

andrixnet:
I can confirm this issue on OCPL, OCNL, OCRO. 
Confirmed across multiple browsers: Firefox (2 versions), Chrome.
Confirmed with addons like Adblock and Privacy badger disabled.

Here is a bit from the browser console in Chrome:
```
GCT.js:179 Uncaught TypeError: Cannot read property 'DataTable' of undefined
    at new GCT (GCT.js:179)
```

Here is a bit as reported by Firefox console:
```

          TypeError: google.visualization is undefined[Learn More]  GCT.js:179:5
	GCT https://www.opencaching.ro/js/GCT.js:179:5
	<anonymous> https://www.opencaching.ro/search.php:285:15
...
[...]

PR #2175, opened 2020-06-18
#2164 workaround, partially tested

rapotek:
Did not want to create pull requests any more (too much stress for me i.a.) but it seems to be an emergency, so this one can make a little improvement until there will be something more acceptable.
It is partially tested:

- checked search results and some stats, did not check merit badges because I do not have any defined in my dev environment,
- works on latest Firefox, did not check on Chrome (should work) or Edge or IE 11 (should be verified).

The idea in short: do any operations of columns definition, row additions etc. on plain objects and arrays until google API is available (loaded), then initialize API with cached data and replace stub methods doing API operations with real ones. The stub is not fully functional but methods necessary to work before API is loaded should be implemented.

Anyone willing can easily verify it working in one's dev environment, it is only one file to replace.

### DIFF
--- a/public/js/GCT.js
+++ b/public/js/GCT.js
@@ -85,7 +85,7 @@ function GCTLoad( ct, lang, disableLoad )
 
 
 
-function GCT( divId ){
+function GCT(divId, disableLoad) {
 
 
     this.divId = divId;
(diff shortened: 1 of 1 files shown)


rapotek:
@tripper1971 has changed GCT.js already. I assume his changes (as the author) have priority, so this PR is not applicable any more.

PR #2182, opened 2020-07-15
Interactive Map

rapotek:
The repository seems to be almost stale, so...

This is a proposition of changes to an existing Dynamic Map, together with corresponding changes is My Neighbourhood. Because of the nature of changes, as perceived from an external user point, I named it an Interactive Map.

IMPORTANT: Because there is a bunch of changes which I consider should be validated by someone, and not only developers but OC Team and advanced users too, the whole set of changes is made to coexist with an official My Neighbourhood, but replacing it after validation should require few changes only, apart from files relocation. To assure this coexistence, some MyNeighbourhood files have been simply copied to another location, while another ones has been only slightly changed.

What is new from a user point of view:

- A popup with marker details, showed by clicking on map marker, allows now to see not only a top-most marker, but also "move" to the next/previous marker being under the exact coordinates of the "clicked one",
- The visibility and order of markers shown on map is consistent with currently set order and visibility of My Neighbourhood "sections" (latest caches, latest logs etc.) and changing the section order or visibility adjusts the map appearance on client side. Ordering: the most top section markers are displayed on top, the second one are next etc.
- The visibility and order changes should apply to currently opened popup contents too.
- If the marker belongs to a named section, the section title is shown in the top part of popup box.
- When the mouse/cursor is moved over a row containing a particular entry, a corresponding marker is "highlighted" on the map and the "highlight" is toned down on mouse/cursor exit.

What is new from a developer point of view (main things):

- All markers within MapModel are now placed within sections. When a marker does not belong to any section, it should be placed within a special, `_DEFAULT_` section.
- Each section has a corresponding layer created in map and the visibility and ordering are layers' operations.
- `DynamicMap` itself (now `InteractiveMap`), its tools etc. have been rewritten to the more objective version, with prototyping where there is a chance for more than one instance creation.
- `InteractiveMapServices` global object should be now an entry point for every external usage.
- Markers have been rewritten to an objective, prototyped version, with inheritance introduced, and grouped within named 'marker family', where each defined family should have its own styling. Only a 'simple' family, using icons "computed" on server side is available now, but an implementation of more advanced 'okapi' icons is in progress.
- A more detailed information is passed with each marker within `markerData`, it is a step to introduce client-side 'okapi' icons too.

Above changes should be available for validation for OCTeam, Advanced Users and System Admins under '/MyNbhInteractive' relative url.

[...]

PR #2185, opened 2020-08-19
Interactive map okapi markers

rapotek:
The description below is copied and modified from recently closed PR #2182 :

This is a proposition of changes to an existing Dynamic Map, together with corresponding changes is My Neighbourhood. Because of the nature of changes, as perceived from an external user point, I named it an Interactive Map.

IMPORTANT: Because there is a bunch of changes which I consider should be validated by someone, and not only developers but OC Team and advanced users too, the whole set of changes is made to coexist with an official My Neighbourhood, but replacing it after validation should require few changes only, apart from files relocation. To assure this coexistence, some MyNeighbourhood files have been simply copied to another location, while another ones has been only slightly changed.

What is new from a user point of view:

- A popup with marker details, showed by clicking on map marker, allows now to see not only a top-most marker, but also "move" to the next/previous marker being under the exact coordinates of the "clicked one",
- The visibility and order of markers shown on map is consistent with currently set order and visibility of My Neighbourhood "sections" (latest caches, latest logs etc.) and changing the section order or visibility adjusts the map appearance on client side. Ordering: the most top section markers are displayed on top, the second one are next etc.
- The visibility and order changes should apply to currently opened popup contents too.
- If the marker belongs to a named section, the section title is shown in the top part of popup box.
- When the mouse/cursor is moved over a row containing a particular entry, a corresponding marker is "highlighted" on the map and the "highlight" is toned down on mouse/cursor exit.
- User can select one of supplied styles for map icons (markers): 'Simple' or 'Okapi' on My Neighbourhood configuration page. The Okapi style is more advanced, with icon size change and caption visibility behavior similar to the main map tiles, generated by OKAPI.

What is new from a developer point of view (main things):

- All markers within MapModel are now placed within sections. When a marker does not belong to any section, it should be placed within a special, `_DEFAULT_` section.
- Each section has a corresponding layer created in map and the visibility and ordering are layers' operations.
- `DynamicMap` itself (now `InteractiveMap`), its tools etc. have been rewritten to the more objective version, with prototyping where there is a chance for more than one instance creation.
- `InteractiveMapServices` global object should be now an entry point for every external usage.
- Markers have been rewritten to an objective, prototyped version, with inheritance introduced, and grouped within named 'marker family', where each defined family should have its own styling. A 'simple' family is using icons "computed" on server side while 'okapi' family determines current icon source basing on cache and log data passed with each marker within `markerData`.
- The information passed with each marker is much more detailed, so the marker icon style computation can be more precised.
- There is a lot of new image files for 'okapi' style, in `public/images/map_markers/okapi` folder, where in most cases one file maps one cache or log detailed state. The 'new' suffixed icons, defined in OKAPI but not used currently, are created too, to be compliant with full OKAPI style. I suppose, they might be useful in the future.
- The 'okapi' style adds additional, 'background' layer with zoom dependant transparency, in a manner similar to the tiles generated by OKAPI.

Above changes should be available for validation for OCTeam, Advanced Users and System Admins under '/MyNbhInteractive' relative url.

Should work (tested) on Firefox 78, Chrome 83, Edge 44. A 'Simple' style does not show latest logs on IE 11 because of SVG icons, but current My Neighbourhood does not show them too. An 'Okapi' style should work (tested) on IE 11.


kojoty:
HI @rapotek 

what's the status of this PR? Is this ready to test?

rapotek:
It should be, since 19th of August. But I have to recall all details coded here.

kojoty:
OK, so let me test and review what's going here..

BTW: if something works too slowly here please feel free to scream loudly :) - at least I have less time and energy for OC sometimes but usually if someone request something I try to meet the requests

kojoty:
@rapotek i tried to check how it works in my local env. but I'm not sure how to see such new map. Is it integrated with any existed view? If not can you just fill the example in test controller...

rapotek:
@kojoty It is written in the OP, but in short:

- Log in as a OC Team member, System Admin or Advanced User
- Go to `/MyNbhInteractive` page in your local env. Most of changes are visible on this page, where there is a changed map integrated with modified MyNeighbourhood controller.

If you want to see OKAPI-based style markers (more advanced than the simple ones), set a corresponding option on the main neighbourhood configuration page.

PR #2193, opened 2021-01-08
viewprofile minor fixes

kojoty:


rapotek:
Pulled and verified on my dev environment - works correctly.

PR #2208, opened 2021-01-23
fix of the issue with pic attach to geochache log

kojoty:
this is critical fix as pic attache to log doesn't work


kojoty:
I commit during the weekend as this is critical.

PR #2209, opened 2021-01-24
Fixes for PHP-7.4

kojoty:
Mostly removing $usr var at all.
Next parts will be ready soon.


kojoty:
OK, $usr is removed completely. @deg-pl  do you know what else blocks in php 7.4?

deg-pl:
@kojoty If I remember well, biggest problem was in viewlogs.php and related files.

kojoty:
> @kojoty If I remember well, biggest problem was in viewlogs.php and related files.

I believe I've resolved all issue I had seen and logs works for me on php7.4 - but if you have a free moment please pull my changes and test on your side

rapotek:
I checked some changes in my dev environment (switching to 7.4), not all of them because there is a lot of files, but the most important ones. Each checked seems to work.

I found, however, one another code which should be fixed, I think. `/src/Views/sysAdmin/dbUpdate.tpl.php` uses `$view->routineFiles as $name => $lastRun` where `$lastRun` is referred to as an array but it may be `null` when there are no previous runs of given routine. Later code f.ex. `Formatter::dateTime($lastRun['runTime']), 0, 10)` gives now current datetime in case of `null` but there will be an error in 7.4 version.

kojoty:
> I checked some changes in my dev environment (switching to 7.4), not all of them because there is a lot of files, but the most important ones. Each checked seems to work.
> 
> I found, however, one another code which should be fixed, I think. `/src/Views/sysAdmin/dbUpdate.tpl.php` uses `$view->routineFiles as $name => $lastRun` where `$lastRun` is referred to as an array but it may be `null` when there are no previous runs of given routine. Later code f.ex. `Formatter::dateTime($lastRun['runTime']), 0, 10)` gives now current datetime in case of `null` but there will be an error in 7.4 version.

Thanks Rapotek, I will look at this. I'm going to update it today and would be nice to merge before weekend.

kojoty:
I would like to merge this PR tonight. Any concerns / comments?

PR #2212, opened 2021-02-06
DB update for new OC voting system

kojoty:
Hi, 

I'm preparing the new voting system for OC. This PR is to deliver DB updates with new tables.
This is mainly for OCPL COG election but it can be used also for any other voting/election etc.

Code will be added in separate PR soon.

### DIFF
--- a/src/Utils/Database/Updates/103_new.php
+++ b/src/Utils/Database/Updates/103_new.php
@@ -1,0 +1,83 @@
+<?php
+
+/**
+ * Tables used by internal OC voting system
+ */
+namespace src\Utils\Database\Updates;
+
+class C16126270995966 extends UpdateScript
+{
+    public function getProperties()
+    {
+        return [
+            // see /docs/DbUpdate.md
+            'uuid' => '3835F75B-A46A-26B0-F728-E4A0FC610851',
+            'run' => 'auto',
+        ];
+    }
+
+    // IMPORTANT:
+    // Any output by 'echo', 'print' etc. will be PUBLIC (see #1923).
+    // Do not output any sensitive information.
+
+    public function run()
+    {
+        $this->db->simpleQueries("
+            DROP TABLE IF EXISTS `vote_voters`;
+            DROP TABLE IF EXISTS `vote_votes`;
+            DROP TABLE IF EXISTS `vote_choiceOptions`;
+            DROP TABLE IF EXISTS `vote_elections`;
+
+            CREATE TABLE `vote_elections` (
+              `electionId` int(11) NOT NULL AUTO_INCREMENT COMMENT 'id of the record',
+              `name` text NOT NULL COMMENT 'election name',
+              `startDate` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP COMMENT 'beginning of election ',
+              `endDate` timestamp NOT NULL DEFAULT '0000-00-00 00:00:00' COMMENT 'end of election ',
+              `voterCriteria` text NOT NULL COMMENT 'JSON with criteria for voters',
+              `electionRules` text NOT NULL COMMENT 'JSON with election data',
+              `description` text NOT NULL COMMENT 'description of election to display',
+              PRIMARY KEY (`electionId`)
+            ) ENGINE=InnoDB COMMENT='Voting data';
+
+            CREATE TABLE `vote_choiceOptions` (
+              `optionId` int(11) NOT NULL AUTO_INCREMENT COMMENT 'unique option id',
+              `electionId` int(11) NOT NULL COMMENT 'id of election',
+              `name` text NOT NULL COMMENT 'name of the option',
+              `description` text COMMENT 'description of the option',
+              `link` text COMMENT 'link added to option',
+              `orderIdx` int(11) NOT NULL COMMENT 'order of the option on the list',
+              PRIMARY KEY (`optionId`),
+              KEY `FK_vote_choiceOptions_vote_elections` (`electionId`),
+              CONSTRAINT `FK_vote_choiceOptions_vote_elections` FOREIGN KEY (`electionId`) REFERENCES `vote_elections` (`electionId`) ON DELETE CASCADE ON UPDATE CASCADE
+            ) ENGINE=InnoDB COMMENT='Options available in the voting';
+
+            CREATE TABLE `vote_votes` (
+              `voteId` int(11) NOT NULL AUTO_INCREMENT COMMENT 'unique id of the vote',
+              `electionId` int(11) NOT NULL COMMENT 'id of the election',
+              `optionId` int(11) NOT NULL COMMENT 'option selected in voting ',
+              `date` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP COMMENT 'date of the vote',
+              `hash` text COMMENT 'optional hash to store additional data',
+              PRIMARY KEY (`voteId`),
+              KEY `FK_vote_votes_vote_elections` (`electionId`),
+              KEY `FK_vote_votes_vote_choiceOptions` (`optionId`),
+              CONSTRAINT `FK_vote_votes_vote_choiceOptions` FOREIGN KEY (`optionId`) REFERENCES `vote_choiceOptions` (`optionId`) ON DELETE CASCADE ON UPDATE CASCADE,
+              CONSTRAINT `FK_vote_votes_vote_elections` FOREIGN KEY (`electionId`) REFERENCES `vote_elections` (`electionId`) ON DELETE CASCADE ON UPDATE CASCADE
+            ) ENGINE=InnoDB COMMENT='Votes from elections';
+
+            CREATE TABLE `vote_voters` (
+              `userId` int(11) NOT NULL COMMENT 'id of the user (from user table)',
+              `electionId` int(11) NOT NULL COMMENT 'id of election (from vote_elections)',
+              `ip` varchar(15) NOT NULL COMMENT 'IP address of the user',
+              `additionalData` text NOT NULL COMMENT 'Additional data for user verification',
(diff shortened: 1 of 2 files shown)


rapotek:
The tables looks like the are related to one another, by `electionId` for example. Shouldn't there be foreign keys defined?

By the way: I am personally glad that an election system enters OCPL code, but I remember when I suggested to introduce some functionality related to election, I was told definitely that the election is an inner Polish matter and OCPL code should not be involved into it in any way. Has anything changed in that?

kojoty:
> The tables looks like the are related to one another, by `electionId` for example. Shouldn't there be foreign keys defined?

I've thought, maybe I will add it.

> By the way: I am personally glad that an election system enters OCPL code, but I remember when I suggested to introduce some functionality related to election, I was told definitely that the election is an inner Polish matter and OCPL code should not be involved into it in any way. Has anything changed in that?

I will answer you @ priv.

deg-pl:
> The tables looks like the are related to one another, by `electionId` for example. Shouldn't there be foreign keys defined?
YYYEEESSSSS!!!!

kojoty:
Foreign keys are there now...

rapotek:
I looked a little bit closer at it yesterday and have some remarks and questions:

- Unless all defined DDLs are run in a single transaction, `vote_elections` table should be defined before `vote_choiceOptions` because the latter use `electionId` as a FK.
- I didn't use a new dbUpdater in means of defining something, so I'm not sure if construction used is required by it, but there are too many alters for me. The same result can be achieved using f.ex.:
```
CREATE TABLE `vote_elections` (
`electionId` int(11) NOT NULL AUTO_INCREMENT COMMENT 'id of the record',
`name` text NOT NULL COMMENT 'election name',
`startDate` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP COMMENT 'beginning of election ',
`endDate` timestamp NOT NULL DEFAULT '0000-00-00 00:00:00' COMMENT 'end of election ',
`voterCriteria` text NOT NULL COMMENT 'JSON with criteria for voters',
`electionRules` text NOT NULL COMMENT 'JSON with election data',
`description` text NOT NULL COMMENT 'description of election to display',
PRIMARY KEY (`electionId`)
) ENGINE=InnoDB COMMENT='Voting data';

CREATE TABLE `vote_choiceOptions`(
`optionId` int(11) NOT NULL AUTO_INCREMENT COMMENT 'unique option id',
`electionId` int(11) NOT NULL COMMENT 'id of election',
`name` text NOT NULL COMMENT 'name of the option',
`description` text COMMENT 'description of the option',
`link` text COMMENT 'link added to option',
`orderIdx` int(11) NOT NULL COMMENT 'order of the option on the list',
PRIMARY KEY (`optionId`),
CONSTRAINT `FK_vote_choiceOptions_vote_elections` FOREIGN KEY (`electionId`) REFERENCES `vote_elections` (`electionId`) ON DELETE CASCADE
) ENGINE=InnoDB COMMENT='Options available in the voting';

CREATE TABLE `vote_voters` (
`userId` int(11) NOT NULL COMMENT 'id of the user (from user table)',
`electionId` int(11) NOT NULL COMMENT 'id of election (from vote_elections)',
PRIMARY KEY (`userId`,`electionId`),
CONSTRAINT `FK_vote_voters_vote_elections` FOREIGN KEY (`electionId`) REFERENCES `vote_elections` (`electionId`) ON DELETE CASCADE
) ENGINE=InnoDB COMMENT='User participation in votings';

CREATE TABLE `vote_votes` (
`voteId` int(11) NOT NULL AUTO_INCREMENT COMMENT 'unique id of the vote',
`electionId` int(11) NOT NULL COMMENT 'id of the election',
`optionId` int(11) NOT NULL COMMENT 'option selected in voting ',
`date` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP COMMENT 'date of the vote',
`hash` text COLLATE utf8mb4_unicode_ci COMMENT 'optional hash to store additional data',
PRIMARY KEY (`voteId`),
CONSTRAINT `FK_vote_votes_vote_elections` FOREIGN KEY (`electionId`) REFERENCES `vote_elections` (`electionId`) ON DELETE CASCADE,
CONSTRAINT `FK_vote_votes_vote_choiceOptions` FOREIGN KEY (`optionId`) REFERENCES `vote_choiceOptions` (`optionId`) ON DELETE CASCADE
) ENGINE=InnoDB COMMENT='Votes from elections';
```
which is much simpler IMHO.
- I think there should be a FK from `vote_voters` table to `user`, while there is a `userId` defined
- I would advise to add `ON UPDATE CASCADE` rule to FKs. MySQL manual says that in case of missing rule default action is 'RESTRICT'
- Is camelCase naming for table names and columns a recommended standard for OCPL DB?
- Looking at the tables relationship chart below, there is a cycle, and therefore I ask: is `electionId` in `vote_votes` necessary while it can be derived from `vote_choiceOptions`?
votes

kojoty:
> I looked a little bit closer at it yesterday and have some remarks and questions:

Thanks!
 
> * Unless all defined DDLs are run in a single transaction, `vote_elections` table should be defined before `vote_choiceOptions` because the latter use `electionId` as a FK.

I think if FKJ are defined in alters this is OK.

> * I didn't use a new dbUpdater in means of defining something, so I'm not sure if construction used is required by it, but there are too many alters for me. The same result can be achieved using f.ex.: 

I export structure by phpmyadmin - that's why it looks like that - I believe this is OK - SQLs works for me - I've tested.

> * I think there should be a FK from `vote_voters` table to `user`, while there is a `userId` defined

I had some mixed feeling about this but you're right - I will add also this FK..

> * I would advise to add `ON UPDATE CASCADE` rule to FKs. MySQL manual says that in case of missing rule default action is 'RESTRICT'

OK, agree - let's do that.

> * Is camelCase naming for table names and columns a recommended standard for OCPL DB?

Yes, I think camelcase is the target approach + prefixes with "_" like: vote_userVotes => vote is prefix to order these tables together. 

> * Looking at the tables relationship chart below, there is a cycle, and therefore I ask: is `electionId` in `vote_votes` necessary while it can be derived from `vote_choiceOptions`?

Yes, but is the cycle a problem?
I know this makes a data redundancy but also it allows to construct simpler data queries - I think this is accepted - isn't it?

rapotek:
> I export structure by phpmyadmin - that's why it looks like that - I believe this is OK - SQLs works for me - I've tested.

I assume it works. I don't use phpmyadmin at all, so I prefer to use more compact script, like an output from 'show create table' command in mysql, it is much more readable for me.

> Yes, I think camelcase is the target approach + prefixes with "_" like: vote_userVotes => vote is prefix to order these tables together.

I have some mixed feelings about this, because I always assumed case sensitivity should be avoided in case-insensitive SQL names, especially when underlying operating system allows to make a 'SampleTable' and 'sampletable' as two different tables (dependant on settings).

> I know this makes a data redundancy but also it allows to construct simpler data queries - I think this is accepted - isn't it?

As far as I know, cycles should be avoided whenever possible, but I am not an expert in this matter.

deg-pl:
> I have some mixed feelings about this, because I always assumed case sensitivity should be avoided in case-insensitive SQL names, especially when underlying operating system allows to make a 'SampleTable' and 'sampletable' as two different tables (dependant on settings).
Yes! MariaDB for Windows assumes that 'TablE' and 'table' - are the same tables. In MariaDB for Linux these are two different tables!

src/Utils/Database/Updates/103_new.php
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    // IMPORTANT:
+    // Any output by 'echo', 'print' etc. will be PUBLIC (see #1923).
+    // Do not output any sensitive information.
+
+    public function run()
+    {
+        $this->db->simpleQuery("
+DROP TABLE IF EXISTS `vote_votes`;
```
rapotek, line 26: In case of FKs included into `CREATE TABLE`, the order should be for example:

- ``DROP TABLE IF EXISTS `vote_votes`, `vote_voters`, `vote_choiceOptions`, `vote_elections` `` (\`vote_votes\` before \`vote_choiceOptions\` and \`vote_elections\`, \`vote_voters\` before \`vote_elections\`, \`vote_choiceOptions\` before \`vote_elections\`)
- `CREATE TABLE` \`vote_elections\`, then \`vote_choiceOptions\`, then \`vote_voters\` and \`vote_votes\` (\`vote_voters\` may be earlier but not before \'vote_elections\`)

This way tables would be correctly dropped and created regardless of transactions etc.

Anyway, I like this way of DDL coding, it is much more readable for me :)

src/Utils/Database/Updates/103_new.php
```diff
@@ -0,0 +1,83 @@
...
+            CREATE TABLE `vote_voters` (
+              `userId` int(11) NOT NULL COMMENT 'id of the user (from user table)',
+              `electionId` int(11) NOT NULL COMMENT 'id of election (from vote_elections)',
+              `ip` varchar(15) NOT NULL COMMENT 'IP address of the user',
+              `additionalData` text NOT NULL COMMENT 'Additional data for user verification',
+              PRIMARY KEY (`userId`,`electionId`),
+              KEY `FK_vote_voters_vote_elections` (`electionId`),
+              CONSTRAINT `FK_vote_voters_user` FOREIGN KEY (`userId`) REFERENCES `user` (`user_id`) ON DELETE NO ACTION ON UPDATE CASCADE,
+              CONSTRAINT `FK_vote_voters_vote_elections` FOREIGN KEY (`electionId`) REFERENCES `vote_elections` (`electionId`) ON DELETE CASCADE ON UPDATE CASCADE
+            ) ENGINE=InnoDB COMMENT='User participation in votings';
```
rapotek, line 76: I think this will do :)

kojoty:
I would like to merge this PR tonight. Any concerns/comments?

PR #2215, opened 2021-02-21
OcConfig cleanup

jrmajor:
OcConfig classes and files in config/ have been formatted according to PSR-12. Docblocks have been normalized, redundant ones removed. Some type hints have been added.

Except two misspelled methods, which were renamed, no public functionality has been changed.


src/Models/OcConfig/EmailConfigTrait.php
```diff
@@ -1,161 +1,127 @@
...
     {
-        $email = self::getEmailAddrVar('ocTeamContactEmail');
-        if($forWebDisplay){
-            return self::emailToDisplay($email);
-        } else {
-            return $email;
-        }
+        $email = self::getEmailAddressFromConfig('ocTeamContactEmail');
+
+        return $forWebDisplay
```
kojoty, line 26: This is minor thing: I have mixed feelings for this format - I feel I don't like this... but it can stay as is - this is only my own preference.

jrmajor: I'm not sure what format you mean, ternary operator?

kojoty: Please note I'm not sure of my opinion in this case

What I mean:
- code readable is a priority

- I prefer not to use new lines when it's not necessary - it makes code block artificially longer - but this is only general approach - lots of exceptions could be possible here

- I think the ternary operator is OK when you really need to do something in one line - I think generally this is less readable - but maybe in this case it could be OK

- terany operator with new lines looks strange in my eyes - but maybe this is only my feeling...

If you like this construction please leave this as it is - I'm OK with this.

jrmajor: OK, I will leave it like this for now, we may discuss it tomorrow :)

deg-pl:
@jrmajor I'm nearly sure, you use PHPStorm ;)

src/Models/OcConfig/ConfigReader.php
```diff
@@ -40,180 +35,120 @@ abstract class ConfigReader
...
-        if(is_file($localMenuFile)){
-            include($localMenuFile);
-            return $menu;
-        }
-
         $ocNode = self::getOcNode();
-        $nodeMenuFile = self::MENU_DIR."$menuPrefix.$ocNode.php";
-        if(is_file($nodeMenuFile)){
-            include($nodeMenuFile);
+        $localMenuFile = self::MENU_DIR . "{$menuPrefix}.local.php";
```
kojoty, line 56: what is a difference between:
`"{$menuPrefix}.local.php"`
and
`"$menuPrefix.local.php"`

src/Models/OcConfig/ConfigReader.php
```diff
@@ -40,180 +35,120 @@ abstract class ConfigReader
...
-            include($defaultMenuFile);
+        if (is_file($nodeMenuFile)) {
+            include $nodeMenuFile;
+
             return $menu;
         }
 
-        Debug::errorLog("ERROR: Can't load menu file: $menuPrefix");
-        return null;
+        include self::MENU_DIR . "{$menuPrefix}.default.php";
```
kojoty, line 71: just a question: why do you removed:
`if(is_file($defaultMenuFile)){`
and logging an error if there is no such file?

src/Models/OcConfig/ConfigReader.php
```diff
@@ -40,180 +35,120 @@ abstract class ConfigReader
...
 
         return $ctrl->links;
     }
 
     /**
-     * Return $config merged from:
-     * - default config file overrirded by
-     * - node-config file overrided by
-     * - local config file
+     * Get config merged from default, node-specific and local config files.
```
kojoty, line 88: this is not much important but I still think that previous comment here was better :)

src/Models/OcConfig/OcConfig.php
```diff
@@ -48,60 +53,102 @@ final class OcConfig extends ConfigReader
...
     }
 
     /**
-     * Private ctor so nobody else can instance it
+     * Private constructor so nobody else can instantiate it.
      */
     protected function __construct()
     {
-        parent::__construct();
         $this->loadConfig();
```
kojoty, line 145: why parent constructor call is removed?
i think every class which extends any other should by default call parent constr.

someone can add constructor to parent and forget to do it for all child class...

kojoty:
OK, for me it looks quite well.
Thanks @jrmajor

src/Models/OcConfig/ConfigReader.php
jrmajor: In this case there is no difference between these, but I think it is better to explicitly state whether string interpolation is  intended. For example `"$this->method()"` and `"{$this->method()}"` would work differently. I think we should decide which syntax we want to use in this project and stick to it. It is my personal preference to:
- Always use `{}` for string interpolations,
- Use `""` for strings, where string interpolation is used or `'` character is used and it would need to be escaped,
- Use `''` for every other string.

I thought there is no standard regarding this in this project, so I used my preferred syntax. For me, however, consistency is more important than personal preferences, so I would by happy to comply with project standards if they exist or if you decide that there is no need to use `{}`.

src/Models/OcConfig/OcConfig.php
jrmajor: I think calling empty parent constructors or defining them only to call them in extending classes is unnecessary and introduces excessive clutter. We have a few reviewers to catch such mistakes ;)

Do you want me to restore this call?

kojoty: no, let it be as it is. 
Empty constructor is not needed. 
I only think that if child class has constructor it can call parent constructor by default - even if the parent class doesn't have an constructor - but I think it's rather a good practice - not a bug

src/Models/OcConfig/ConfigReader.php
jrmajor: For sure it was more detailed, but I think it's pretty straightforward what it does. I prefer short docblocks with only necessary information. If somebody uses this method, he will not care about configuration merging logic, and if he wants to modify it, he will read it anyways.

src/Models/OcConfig/ConfigReader.php
jrmajor: Node specific and local config files may or may not exist, but the default one should always be available as a fallback (except site.*, which contains sensitive information, like DB credentials). I should've replaced `include` with `require` to make sure an error will be thrown when it is missing.

PR #2223, opened 2021-02-25
I18n cleanup

jrmajor:
Cleanup like in #2215. Also a bug has been fixed in MainLayoutController.

### DIFF
--- a/src/Utils/I18n/I18n.php
+++ b/src/Utils/I18n/I18n.php
@@ -14,14 +14,10 @@ namespace
      * Translate the given message.
      *
      * @param string $key The message key in /lib/languages/*.php
-     * @param array|null $replace Replacements to be made using vsprintf()
+     * @param array $replace Replacements to be made using vsprintf()
      */
-    function tr(string $key, array $replace = null): string
+    function tr(string $key, array $replace = []): string
     {
-        if (! $replace) {
-            return I18n::translatePhrase($key);
-        }
-
         return vsprintf(I18n::translatePhrase($key), $replace);
     }
 }
--- a/src/Utils/I18n/I18n.php
+++ b/src/Utils/I18n/I18n.php
@@ -1,418 +1,403 @@
 <?php
 
 /**
- * Global nameespace here is only to define global function tr
- * used to translate phrases in OC code
+ * Global namespace is here to define the global tr() function
+ * used to translate phrases in OC code.
  *
- * After refactoring all views it should be moved to View class.
+ * After refactoring all views it should be moved to the View class.
  */
-namespace {
-
+namespace
+{
     use src\Utils\I18n\I18n;
 
     /**
-     * Return tranlated string
-     *
-     * @param string $str - translation key
-     * @param array $args - arguments to insert into string (see vsprintf for details)
-     * @return string - localized string
+     * Translate the given message.
      */
-    function tr($str, array $args = null)
+    function tr(string $message, array $replace = null): string
     {
-        if (is_null($args)) {
-            return I18n::translatePhrase($str);
-        } else {
-            return vsprintf(I18n::translatePhrase($str), $args);
+        if (! $replace) {
+            return I18n::translatePhrase($message);
         }
 
+        return vsprintf(I18n::translatePhrase($message), $replace);
     }
+}
 
-} // namespace / (global)
-
-namespace src\Utils\I18n {
-
+namespace src\Utils\I18n
+{
     use Exception;
-    use src\Utils\Text\UserInputFilter;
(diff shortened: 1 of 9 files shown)


src/Utils/I18n/I18n.php
```diff
@@ -1,418 +1,403 @@
...
-namespace {
-
+namespace
+{
     use src\Utils\I18n\I18n;
 
     /**
-     * Return tranlated string
-     *
-     * @param string $str - translation key
```
kojoty, line 16: Moim zdaniem ten komentarz ma sens - postuluję zostawić go lub dodać coś innego, bardziej opisowego

src/Utils/I18n/I18n.php
```diff
@@ -1,418 +1,403 @@
...
-    use src\Utils\Debug\Debug;
-    use src\Utils\Database\XDb;
     use src\Utils\Cache\OcMemCache;
+    use src\Utils\Database\XDb;
+    use src\Utils\Debug\Debug;
+    use src\Utils\Uri\OcCookie;
+    use src\Utils\Uri\Uri;
 
-    class I18n
+    final class I18n
```
kojoty, line 36: why "final"?

src/Utils/I18n/I18n.php
```diff
@@ -1,418 +1,403 @@
...
+
+            if (! $instance) {
+                $instance = (new static())->initialize();
             }
 
             return $instance;
         }
 
-        private function initialize()
+        protected function initialize(): self
```
kojoty, line 59: final class and protected now?

src/Utils/I18n/I18n.php
```diff
@@ -1,418 +1,403 @@
...
-        public static function translatePhrase($translationId, $langCode=null, $skipPostprocess=false, $skipFailoverLang=false)
-        {
-            return self::instance()->translate($translationId, $langCode, $skipPostprocess, $skipFailoverLang);
+        public static function translatePhrase(
+            string $message,
+            string $language = null,
+            bool $skipPostprocess = false,
+            bool $skipFallback = false
+        ): string {
+            return self::instance()->translate(...func_get_args());
```
kojoty, line 129: hmm... I'm not a supporter of such tricks - I think much more readable is to have the list of args

src/Utils/I18n/I18n.php
```diff
@@ -1,418 +1,403 @@
...
 
         /**
-         * Allow to check if given phrase is present
-         *
-         * @param string $str - phrase to translate
-         * @return boolean
+         * Check if the given message exists.
          */
-        public static function isTranslationAvailable($str)
+        public static function isTranslationAvailable(string $message): bool
```
kojoty, line 135: minor thing but "message" is rather not a good name for a translation key...

src/Utils/I18n/I18n.php
```diff
@@ -1,418 +1,403 @@
...
+                $translations[substr($key, $prefixLength)] = $this->trArray[$currentLanguage][$key] ?? $text;
             }
 
-            return $t;
+            return $translations;
         }
 
-
         /**
-         * Methods for retrieving and maintaining old-style database translations.
```
kojoty, line 363: please leave this comment - I think it explains what is going on here

src/Utils/I18n/I18n.php
rapotek: Accoring to documentation, array spread operator was introduced in 7.4. While there is much work done in adjusting code to 7.4, this operator should not be used until all instances are running on 7.4 or higher versions, IMHO. Are they running in PHP 7.4 by now?

kojoty:
OK, Review is done. Everything looks nice. We can go with it (if it's ready).

src/Utils/I18n/I18n.php
jrmajor: @kojoty These methods have identical signatures, which are pretty long, so I think it is more readable and makes it clear that this is just an alias for a private method. I'll revert it if you don't like it.

@rapotek It's argument unpacking, introduced in PHP 5.6.

rapotek: > @rapotek It's argument unpacking, introduced in PHP 5.6.

Sorry for incorrect interpretation.

jrmajor:
I want to move the `tr()` function to another file with global helpers only, and was waiting for #2213 to be merged to autoload this file, but I can open another PR for that.

src/Utils/I18n/I18n.php
jrmajor: It seems cleaner to me. Do you want me to revert it?

PR #2231, opened 2021-03-08
Install  PHP-CS-Fixer 

jrmajor:
This PR installs [PHP-CS-Fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer) and applies its fixes.

I can change the configuration if any adjustments are required. Code style is subject to discussion.


jrmajor:
Sorry, missclick. This PR requires further work.

kojoty:
@jrmajor Could you please explain in a few words what this PR contains.
I assume there is: 
1. installation of cs-fixer by composer in /vendor
2. config specific for OC 
3. code style fixes in many files - I assume it is done automatically by cs-fixer

If so can we split this PR to two parts:
- first part 1.+2. 
- second part 3.
There are so many changes that it's almost impossible to review.

deg-pl:
@jrmajor please don't convert:
`$string = 'a' . 'b';
`
to
`$string = 'a'.'b';
`

From PSR-12 point 6.2:

> All binary arithmetic, comparison, assignment, bitwise, logical, string, and type operators MUST be preceded and followed by at least one space

jrmajor:
@kojoty 

CS-Fixer is installed via Composer. In the first commit manual changes have been made in `composer.json`, `.gignore` (ignore `.php_cs.cache`) and `.php_cs` has been added. All the rest has been done by Composer. `.php_cs` contains configuration.

The second commit indeed contains automatic fixes. I will leave it for now as a preview of code style enforced by this config and revert it when we want to merge.

@deg-pl OK, I will change it.

kojoty:
OK, so what is a purpose of php-cs-fixer on the production server?
I understand I can use it in my dev . env to format code - right? 
We can even share a config to use the same format, but what is a plan to have this in the repo on each server?

jrmajor:
@kojoty It's added as a dev-dependency. If we did not keep `/vendor` in git, it wouldn't be installed on production servers. It may be used to format code in dev environment (by running `vendor/bin/php-cs-fixer fix`). If you're willing to enable Actions for this repo, I would also add code style checks style on pull requests.

kojoty:
OK, so if the php-cs-fixer is only to run in dev. env. why we need it in production repo?

Please correct me if i understand this wrong but anyone can install it outside of the dir with OC code and run automatic code styler on the OC code manually - ocfixer gets the source files location as an input param... Then I think we can store config for ocfixer in docs dir - the rest of files is not needed.

Running it automatically for each commit/PR is another thing - we can think about it but I'm not a big fun of this idea.
Do you think about running cs-fixer as a hook in git locally and run on commit or integrate it with github some way?

kojoty:
BTW: running cs-fixer locally and committing the files with fixed code style from time to time could be definitely a good idea.

kojoty:
> > OK, so if the php-cs-fixer is only to run in dev. env. why we need it in production repo?
> 
> We do not need it there. In `composer.json` it is listed in `require-dev`, which should contain dependencies used only in local environment (test frameworks, static code analysers, linters), and CS-Fixer fits this category. If we didn't keep `/vendor` in git, in production we would run `composer install --no-dev`, which would skip these dependencies. However, as we do that, I think keeping only production dependencies in git is not a good idea. Installing dev dependencies locally would probably raise conflicts even when checking out a different branch, and you would need to take care of excluding changes in `/vendor` from commits (bear in mind that it's not the only dev dependency, kint should be also moved to this category after we remove all references to it from the source code).

OK, I still think the cs-fixer is an optional tool used only on dev. env so there is no need to install it by composer with OC source
You can always has this installed by composer but in different directory than OC source and OC /vendor dir (like a separate project installed by composer) and run this on OC code pointed the right dir and config in folder with OC code.


> > Then I think we can store config for ocfixer in docs dir
> 
> `.php_cs` file in project root directory is a default configuration file location for PHP-CS-Fixer. When we finally move all publicly available files to `/public`, clutter in root directory will be reduced and `.php_cs` won't be even available from a browser. I think it should stay there.

I believe the key word here is "when" - "when" can be in year or a few years - we have been working on this for years now :)

> 
> > Do you think about running cs-fixer as a hook in git locally and run on commit or integrate it with github some way?
> 
> We can use GitHub Actions for that. It is free for public repositories. It would check style according to our configuration on every pull request or push. In the attached screenshot you can see, how it would look (those “8 checks” are GitHub Actions).
> 
> 

**And this seems to be a good point**
Let's try this. I've found quite nice project: https://github.com/OskarStark/php-cs-fixer-ga
Please review this - it seems we don't need to have CS-fixer in the repo and run it by git actions - for me looks very nice - what do you think?

kojoty:
BTW: thanks to the fact that this thread got my attention to github actions I even think it would be nice to migrate to https://github.com/crowdin/github-action from current solution based on github-hooks..

It looks that it can resolve our main issue in dev. work with translations: manually provide the translations in php files - I mean now there is no way to for example rename the translation key without loss of all translation for this key...

jrmajor:
@kojoty

OK, we can install CS-Fixer globally (`composer global require friendsofphp/php-cs-fixer:^2.18`).
You may then run `composer global exec php-cs-fixer fix` in project directory to apply fixes.

> Let's try this. I've found quite nice project: https://github.com/OskarStark/php-cs-fixer-ga

It seems redundant to me. I've set it up without any custom actions. You can see it running here: https://github.com/jrmajor/opencaching-pl/runs/2075602641.

> > > Then I think we can store config for ocfixer in docs dir
> >
> > .php_cs file in project root directory is a default configuration file location for PHP-CS-Fixer. When we finally move all publicly available files to /public, clutter in root directory will be reduced and .php_cs won't be even available from a browser. I think it should stay there.
> 
> I believe the key word here is "when" - "when" can be in year or a few years - we have been working on this for years now :)

It is out of scope for this PR, but maybe the project would be cleaner if we moved publicly available files from root dir to `/public`? From what I understand, we're aiming to eventually do it anyway, I just wonder if it could be done even before rewriting all legacy PHP files.

> I even think it would be nice to migrate to https://github.com/crowdin/github-action from current solution based on github-hooks..

Sounds like a good idea :)

jrmajor:
@kojoty 

> this is docker container ready to run with the cs-fixer inside - so there is no need to install cs-fixer by composer in ubuntu image every time by composer... - I guess thanks to that it can work much faster.

I don't think so — it took 4 seconds to install CS-Fixer, and it gives us the benefit of having control over which versions we use. PHP-CS-Fixer applies rules according to PHP version that is running, so using PHP 7.2 ensures us it's fixing only things, that production servers support (setup-php action allows us to use whichever PHP version we want to).

It's PHP-CS-Fixer run what takes longest (more than 2 minutes) and I don't think it can't be sped up.

Installing PHP took 30 seconds, but I also use this action in my personal projects, and it takes 1-2 seconds, which makes me think it may just need to warm up.

kojoty:
OK. Yes, it allows us to control everything - that's correct - but you need also to maintain it yourself. 

But OK. Let's continue along your way.

.php_cs
```diff
@@ -0,0 +1,186 @@
...
+    ->in(__DIR__)
+    ->name('*.php')
+    ->ignoreDotFiles(true)
+    ->ignoreVCS(true);
+
+return (new PhpCsFixer\Config())
+    ->setRules([
+        '@PSR12' => true,
+
+        // Alias
```
keradus, line 17: you enable a lot of rules.
consider using `@Symfony` or `@PhpCsFixer` ruleset

kojoty:
@jrmajor is this ready to merge?

jrmajor:
@kojoty Yes.

PR #2237, opened 2021-03-10
Update PHP dependencies

jrmajor:
Restores caret version ranges in `composer.json`, as discussed in #2213.

The following packages have been updated:
- HTML Purifier to the next minor version,
- Kint from v1 to v3.

`ddd()` helper has been removed from Kint, so I've replaced it with `exit` in `getLogEntries.php` and `src/Models/PowerTrail/Log.php`. Except manual changes in those files and `composer.json`, all the rest has been done by Composer. As `/vendor` diff can not be reviewed, you may want to run `composer install` to validate that.

I've also added `ext-mysqli` to dependencies list, as it is used by this project.


kojoty:
@jrmajor is this ready to merge?

jrmajor:
@kojoty Yes.

kojoty:
I will try to finally review and merge it today...

Issue #2218, opened 2021-02-23
Better separation between "Page" and "API" controllers - discussion


rapotek:
I would like to propose a small change in controller design approach:

- For every new functionality (or a refactored one) create, if needed, separate controllers for "Page" (HTML code) creation and "API" (json REST API) response,
- API controllers should inherit from `BaseAPIController`, which inherits from `BaseController` and contains methods common for json-based request and response,
- Every API controller should have a name ending with `APIController` suffix, this way f.ex. they could be recognized by `SimpleRouter`, for doing some preprocessing maybe,
- Abandon using `ajax` prefixed methods, use well-defined REST API methods instead.

I think that introducing this approach could make moving to a single-page app in future much simpler.

kojoty:
I'm OK with this proposal but let me understand first this:

>> I think that introducing this approach could make moving to a single-page app in future much simpler.

I'm not web dev so sometimes I can't understand even the most obvious facts :)
I know what is single-page app buty why API methods in separated controllers helps here in anything
(and BTW I think it's quite easy task to move methods form one controller to another in current code - we only need to fix references in the whole code after that)

rapotek:
> I'm OK with this proposal but let me understand first this:
> 
> > > I think that introducing this approach could make moving to a single-page app in future much simpler.
> 
> I'm not web dev so sometimes I can't understand even the most obvious facts :)
> I know what is single-page app buty why API methods in separated controllers help here in anything
> (and BTW I think it's quite easy task to move methods form one controller to another in current code - we only need to fix references in the whole code after that)
> 
> I assume you want to have separate API called by ajax and the rest of controllers - right?

Right, I would like to separate API and the rest of controllers - it will help in a general view how much functionality is left in an existing (page) way and how much is already implemented in a REST API way, IMHO. Think about it like having two different liquors and pouring one from a jar to another rather than mixing two in a single jar ;)
But that's only my view...

PR #2238, opened 2021-03-14
common.inc.php main init script +  controllers refactor

kojoty:
There are several number of changes - mostly cleanups. Main new features:
- new parent controllers ApiBaseController and ViewBaseController (see #2218 for details of the separated controllers idea)
- cleanup of common.inc.php - this script is still use as an entrypoint to many scripts but finally it can be reintegrated to the only entry point script: index.php


src/Controllers/Core/ApiBaseController.php
```diff
@@ -0,0 +1,100 @@
...
+    const HTTP_STATUS_CONFLICT = 409;
+    const HTTP_STATUS_INTERNAL_ERROR = 500;
+
+    /**
+     * Every ctrl should have index method
+     * which is called by router as a default action
+     */
+    public function index()
+    {
+        // API requests must be strict and method mus be present
```
deg-pl, line 24: mus -> must

deg-pl:
Very good job! Thanks!

PR #2245, opened 2021-03-25
php-cs-fixer: change working mode and add diff format

rapotek:
Currently a `php-cs-fixer` action checks every php file in repository, with exclusion of some predefined directories in `.php_cs`. While it is desirable to have a whole code compliant with PSR-12 etc., the experience shows that it won't be fulfilled maybe in years. Therefore I would like to change the action behaviour to check only `*.php` files added or modified in the push or pull request that triggered the action.
In addition the `diff-format=udiff` argument will be added to show line numbers in `php-cs-fixer` diff output. When the 3.0 version will be available, this argument most probably can be safely removed.

The proposed changes uses `dorny/paths-filter@v2` because from all tested solutions it worked the best.

The tests included creating new branch with several commits in one push, including adding, deleting and modifying files, and subsequent pushes with different commits to an existing branch. The pull-request scenario was not tested, I hope the action will work correctly in this case too.
There is a risk that pushes or pulls with a hundred or more changed `*.php` files will not work, because of too long variable/arguments line. The tests worked with a hundred of files, but each within a relative short path.

### DIFF
--- a/.github/workflows/style.yml
+++ b/.github/workflows/style.yml
@@ -10,29 +10,44 @@ jobs:
   php:
     name: PHP
     runs-on: ubuntu-latest
-
     steps:
-      - uses: actions/checkout@v2
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Get changed files
+        uses: dorny/paths-filter@v2
+        id: filter
+        with:
+          list-files: shell
+          base: ${{ github.ref }}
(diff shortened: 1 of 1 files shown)


jrmajor:
For now this would be a better solution. I tried to apply fixers one by one and manually review changes, but there were a lot of them and I didn't want to cause merge conflicts.

Issue #2226, opened 2021-03-03
OCTeam view to remove user account 


kojoty:
Until now there is no simple way to remove user account on user request and site admins do this manually.

I think we should change this.

First let's add the option for OC-Team:

1. search uses account by email address (user should request account deletion by email assigned to account (to be sure who is requester) 

2. then user account can be removed with options:
- account after the operation will be locked 
- account username will be anonymized (because of personal data) - removed account prefix will be provided by node configuration
- user email will be removed
- additional user settings stored in DB will be removed - there is no sense to store for example list of watched caches for removed accounts
- information about the OC-Team member who remove the account will be saved

3. there is no way to undelete "removed" account

Next I'm going to implement the solution to allow user to delete their account - but this will be defined in another issue.

andrixnet:
With all GDPR implications as well. (right to be forgotten, etc)

andrixnet:
IMO this functionality should exists two-fold: 
- user to "delete/deactivate own account forever"
- admin to "delete/deactivate selected account"

With all due warnings and confirmations.

kojoty:
> IMO this functionality should exists two-fold:
> 
> * user to "delete/deactivate own account forever"
> * admin to "delete/deactivate selected account"
> 
> With all due warnings and confirmations.

Yes - I'm want to implement the second option first and the first later - the first option is a little bit more complicated, but most of the code will be reused.

PR #2246, opened 2021-04-05
OCTeam view to remove user account (#2226)

kojoty:
OC Team has now an option to remove user account by click the button.
  Such account is then marekd as removed and all private data is removed from DB.

### DIFF
--- a/config/news.default.php
+++ b/config/news.default.php
@@ -7,4 +7,6 @@
  * It may be customized in node-specific configuration file.
  */
 
+$news = [];
+
 $news['showOnStartPageForNonLoggedUsers'] = false;
--- a/config/user.default.php
+++ b/config/user.default.php
@@ -1,12 +1,24 @@
 <?php
-
 /**
  * User-related configuration
  *
- * This is a default configuration.
+ * This is a default user configuration.
  * It may be customized in node-specific configuration file.
  */
 
-$user = [
-    'defaultStatpicText' => 'Opencaching',
-];
+$config = [];
+
+/**
+ * Removed user account has anonimized username in format: <removedUserUsernamePrefix>_<date-of-removing>_<userId>
+ */
+$config['removedUserUsernamePrefix'] = "Account_removed";
+
+/**
+ * Text displayed in as description of removed user account - formatted date will be added at the end of the text
+ */
+$config['removedUserDescription'] = "Account removed on user request.";
+
+/**
+ * Default text displayed on statpic images
+ */
+$config['defaultStatpicText'] = 'Opencaching';
--- a/config/user.pl.php
+++ b/config/user.pl.php
@@ -1,0 +1,16 @@
+<?php
+/**
+ * User-related configuration
+ *
+ * Those are configuration overrides for OCPL node only.
+ */
+
+/**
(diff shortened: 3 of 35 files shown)


kojoty:
@harrieklomp @andrixnet
Please note this PR contain a new config entries.

kojoty:
How it works:

it is possible now to remove user account from OCTeam user-search menu. There is a new button in top-right menu: "remove account".
After confirmation:
- user account is marked as removed and permanently locked
- username is changed to <prefix-from-config>_id:<user-id>
- user description is changed to <prefix-from-config> <date of removal> 
- password, home coordinates, data associated with this user account in may other account is also removed
- admin note is added to the user account

kojoty:
I'm going to develop the same possibility for every user to remove its own account (in future).

andrixnet:
Questions: 
What happens to caches posted by that account that are in a state other then "archived" ?

andrixnet:
@kojoty: To which are you referring to? 

$news['showOnStartPageForNonLoggedUsers']
$user = [
    'defaultStatpicText' => 'Opencaching',
];
$config['removedUserUsernamePrefix'] = "Account_removed";
$config['removedUserDescription'] = "Account removed on user request.";
$config['defaultStatpicText'] = 'Opencaching';

Shouldn't some of these values be strings from the translation subsystem? (and only manually set if very explicitely needed on that installation?)

For example, if viewing OCPL in english, one should see "Account removed on user request." instead of "Konto usunięte" (which should be displayed when viewing the site in polish).

Furthermore, as I understand so far, this PR contains UI for admins. Then the default value for this string is incorrect:
$config['removedUserDescription'] = "Account removed on user request.";
The accound was removed by an administrator. (for whatever reason, only one of which might be 'user request'). 
That particular string value will be valid in the future, when account removal controls will be implemented for the user and the user was the one to perform the removal.

kojoty:
> 
> 
> Questions:
> What happens to caches posted by that account that are in a state other then "archived" ?

I think that nothing.. All logs and caches stays as they are. If geocache will be unmaintained it will be automatically archived. Any concerns will be handled by OC-team on request.

I think it's the best way to handle user resign decision.

kojoty:
> Furthermore, as I understand so far, this PR contains UI for admins. Then the default value for this string is incorrect:
> $config['removedUserDescription'] = "Account removed on user request.";
> The accound was removed by an administrator. (for whatever reason, only one of which might be 'user request').
> That particular string value will be valid in the future, when account removal controls will be implemented for the user and the user was the one to perform the removal.

I assume that OCTeam never remove any account without the user request. If something is account can be only locked - not removed, so if the account is remove it means that user asked for it.

Like I mentioned I'm going to implement option for "self-remove" account for users as well soon.

andrixnet:
Will it be possible for an admin to reassign a non-archived cache of a removed user to another owner? (like an adoption after user was removed) ?

kojoty:
> 
> 
> Will it be possible for an admin to reassign a non-archived cache of a removed user to another owner? (like an adoption after user was removed) ?

hmm.. as far as I remember there is no such option now but it can be always done directly in DB.
I think it can be dependent to OCTeam decision. We can always implement such feature but I think it's a next corner case.
For OCPL I get usually less that 1 request for account remove per week and most of the users has only a few founds and they are not an owners of any geocache.

PR #2249, opened 2021-04-11
Config cleanups

kojoty:


kojoty:
@deg-pl @andrixnet @harrieklomp 

Please review this PR. There are some number of config changes - please be sure this is OK for your nodes - I made the changes based on configs I got some time ago but maybe something was changed in the meantime

I would like to commit it soon.

andrixnet:
Will check it out on behalf of OCRO and OCUS. (and talk to Harrie). and get back to you

andrixnet:
Side question: why do I see (from Github): running PHP-CS-FIXER failed (duration: 47s - longest among alll checks, by far)?

jrmajor:
@andrixnet What do you mean by “longest among all checks”? Were there any other checks, or this run of this check was the longest one so far?

andrixnet:
1 failing check
Style / PHP (pull_request) Failing after 54s — PHP 
Checked all files in 39.251 seconds, 22.000 MB memory used
Error: Process completed with exit code 8.

github

jrmajor:
@andrixnet These “checks” are steps of this workflow. They do not perform any checks, just set up the environment to run PHP-CS-Fixer, which validates code style of files changed in this commit (see #2231 and #2245). Right now most files have some code style violations, so this check always fails. I'm not sure how we can apply fixes without disrupting everybody else workflows, because they would make many changes across all files in this repo.

andrixnet:
Regarding OCRO, the config changes I saw seem ok, with the important mention that the renamed config directives should also be settable in settings.inc.php (overwriting any setting from /config).

Regarding OCNA: I see no mention about it (on per node defaults). I presume I'll have to change some variables in settings.inc.php, right?

kojoty:
> 
> Regarding OCRO, the config changes I saw seem ok, with the important mention that the renamed config directives should also be settable in settings.inc.php (overwriting any setting from /config).

No. settings.inc.php is a legacy config file and I try to remove all the settings from there.
If you need to override some /config settings do the "local" version of the file - the /config has a 3-level structure:

/config/someConfig.default.php  => here we have a global defaults
/config/someConfig.pl.php => here we have overrides to global defaults for "pl" node
/config/someConfig.local.php => here we have overrides to previous two files but strict local for server - the "local" files are not stored in a git repo

and please not to store anything in "local" files - put everything to the config in repo - local files are intended to be only for:
- some temporary overrides to test on production servers etc.
- specific secrets - like keys, passwords etc. which can't be public visible (like db username / password)

andrixnet:
Is there any documentation for /config/* besides code comments?

kojoty:
> 
> Is there any documentation for /config/* besides code comments?

I don't think we should make any documentation for /config besides comments in the code - such docs have a permanent tendency to be outdated.

The config vars are quite simple. Most of them should be OK with defaults.

Every var should has its complete description in comments in the code - if something is not done - please comment it or let me know and then I will update the comments.

I hope this approach is acceptable for everyone.

andrixnet:
IMO configuration directives should be as stable as possible and well documented. 
With all due respect, I opened https://github.com/opencaching/opencaching-pl/issues/2263 in this regard. 

Volatility is only acceptable for the old, outdated and to be ported config items, until this rewrite is completed.

Please let me know how I can help. I'd try writing the WIKI about it (?)

PR #2264, opened 2021-04-16
#1990: fix lack of the first geochache in GPX

kojoty:


kojoty:
I merge  this PR in weekend as this is minor change which fix bug and this is requested by users.

PR #2267, opened 2021-04-17
Categories in news display

kojoty:
This is a new feature requested by OCPL OCTeam: separate, specific news list about cooperation with other institutions.

I would like to:
- introduce categories to news
    -  there will be common category (for all nodes: news - this will contains all news created till now)
    -  there will be possibility to define node-specific categories and display them under separate links in menu 
- allow to attache images to articles


kojoty:
It seems I made all feature I've planned.
Now:
- image can be attached to news
- every news has the list of associated pictures
- there is a way to define node-specific list (categories) of news
- there is a hidden "draft" category to save drafts / templates etc.

Everything is a little bit raw as tinyMce has limited options to deal with images but it seems its working.

Please test and provide any feedback.

PR #2274, opened 2021-04-23
files cleanup: sqlalters archiving

kojoty:


deg-pl:
Maybe better idea is archiving SQLalters by year - i.e. alters_2010.zip, alters_2009.zip etc?

jrmajor:
Why would we keep them in repo? If somebody really needs them (although I don't know what for) they are in git history anyway. Maybe putting a link like the one below in docs would be enough? It would also be easier to look through them (wouldn't require unzipping).

https://github.com/opencaching/opencaching-pl/tree/7c6677837a9d538d9ce516e68cddc9ade9fd4038/sqlAlters

kojoty:
I don't know why anyone needs it but maybe to debug some historical issue - probably no one uses it
This is really small archive and the readme which describe the situation - I think it can stay in this form and it doesn't affect anything :)

PR #2276, opened 2021-04-23
map engine upgrade: openlayers 5.2 -> 6.5

kojoty:


rapotek:
I've been thinking about it from some time. I hope you checked that our crucial map functionalities would work after upgrade.

But why don't you provide a php-cs rules compliant version? There are only a few changes more in one file, see the action output.

PR #2281, opened 2021-05-02
restore native context menu in tinyMce

kojoty:


kojoty:
minor change requested by users - I commit during the weekend

PR #2297, opened 2021-09-25
Next part of changes necessary for reactivation rules notice

kojoty:


rapotek:
It looks like it is not only reactivation rules but a few more things too (cleaning/moving files). However, why don't you try to be compliant with php-cs rules in new or modern-style files (like config ones)? I do not mean the old ones which should be refactored anyway.

Issue #2295, opened 2021-09-19
Geocache reactivation rules defined by author


kojoty:
TBD

kojoty:
A few screenshots:
![image](https://user-images.githubusercontent.com/2961448/134785211-247cb5d2-1989-4ff9-8399-b72508a52839.png)

![image](https://user-images.githubusercontent.com/2961448/134785227-527b2c63-994c-4051-ba61-9aa9217d1f92.png)

PR #2303, opened 2021-10-15
#2295: final commit for reactivationRules implementation

kojoty:


rapotek:
@kojoty I haven't got time to test it before, but I am doing it now and there is a thing. Correct me if I am wrong, but a verification if any of reactivation rule is selected is done on a client side only. I think, when there is a requirement to select one of available options, it should be verified on a server side, too.
And one more thing: no reactivation rule is selected by default and the user creating a cache has to manually select one every time. I suppose it is made on purpose in this way and has been discussed with Polish OcTeam? Or not?

kojoty:
@rapotek 
No, I think there is no need to validate presence of reactivation rules on server side. I think react.rule is some kind of convention. I even leave the solution to select "custom" rules and leave the textfield blank - it also can be selected - but I think this is a decision of the user to intentionally skip reactivation rules. If someone decide to modify the request to send form without reactivation rules I think it is something what should not be prevented.

And yes, no rules are selected by default because user should decide once for each geocache how to deal with reactivations.

rapotek:
> @rapotek No, I think there is no need to validate presence of reactivation rules on server side. I think react.rule is some kind of convention. I even leave the solution to select "custom" rules and leave the textfield blank - it also can be selected - but I think this is a decision of the user to intentionally skip reactivation rules. If someone decide to modify the request to send form without reactivation rules I think it is something what should not be prevented.

I always thought client-side verification is a convenience, server-side one is security, but... as you wish.

PR #2323, opened 2021-11-15
Prevents passing empty lat or lon in a new cache creation

rapotek:
From OCPL forum:

> zdaje się, że chwilowo istnieje jakiś błąd, który pozwala na dodanie skrzynki bez kordów
a to z kolei wywala stronę główną.

It looks like passing out empty coordinates in a new cache creation, allowed until now, caused a crash on the main page. This fix removes a possibility to store empty (null) values as a new cache coordinates.

It is a simple fix (just `if...else` removal), without any php style checking.

But one question arises: why entering zeros (i.e 00.000) in latitude or longitude for a new cache is forbidden while it is allowed in a cache edition? I did not change this functionality but I would like to know its purpose.

### DIFF
--- a/newcache.php
+++ b/newcache.php
@@ -464,71 +464,67 @@ if (isset($_POST['submitform'])) {
     }
 
     // check coordinates
-    if ($lat_h != '' || $lat_min != '') {
-        if (! mb_ereg_match('^[0-9]{1,2}$', $lat_h)) {
+    if (! mb_ereg_match('^[0-9]{1,2}$', $lat_h)) {
+        tpl_set_var('lat_message', $error_coords_not_ok);
+        $lat_h_not_ok = true;
+    } else {
+        if (($lat_h >= 0) && ($lat_h < 90)) {
+            $lat_h_not_ok = false;
+        } else {
             tpl_set_var('lat_message', $error_coords_not_ok);
             $lat_h_not_ok = true;
-        } else {
-            if (($lat_h >= 0) && ($lat_h < 90)) {
-                $lat_h_not_ok = false;
-            } else {
-                tpl_set_var('lat_message', $error_coords_not_ok);
-                $lat_h_not_ok = true;
-            }
         }
+    }
 
-        if (is_numeric($lat_min)) {
-            if (($lat_min >= 0) && ($lat_min < 60)) {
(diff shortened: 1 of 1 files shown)


andrixnet:
Don't all caches by definition have coordinates?

All zeroes, though geographically meaningless, represents valid data.
However IMO should not be default coord fields contents. Fields should
be empty for the user to fill out and with proper input checking.

rapotek:
> Don't all caches by definition have coordinates? All zeroes, though geographically meaningless, represents valid data. However IMO should not be default coord fields contents. Fields should be empty for the user to fill out and with proper input checking.

I agree, that there should not be defaults for coordinates, and it has been in this way for a while. But it is also possible in working code to omit entering coordinates at all, what results in NULL values stored in database for coordinates. This PR keeps empty initial values for coordinate fields but enforces filling them up while creating a new cache.

Forbidden zero values for coordinates is another thing. I do not know its purpose so I left it as is was, but I suspect it is a kind of old protection against invalid data. However, why this "protection" exists in a new cache creation only and in edition you can freely enter zeroes?

kojoty:
Yes, all-zeroes is in fact very less possible as this is located at ocean near Africa but this is a problem with proper handling of zeros in the code - of course if everything is written ok it should not be a problem ;)

rapotek:
> Yes, all-zeroes is in fact very less possible as this is located at ocean near Africa but this is a problem with proper handling of zeros in the code - of course if everything is written ok it should not be a problem ;)

The code does not allow to create a cache with either 0 lat or 0 lon, they are checked independently. You cannot create a cache on the Greenwich meridian as well as on the equator. I wonder why nobody from OCUK ever complained ;).
BTW, posted data are anyway checked in details about numeric, range etc. Maybe there is a specific case which I don't see and it requires this "zero" comparison, but if not, it looks like being too strict.

andrixnet:
Well, longitude 0 0.000 by itself (with various lat values) is a
perfectly ok value.
The simple check that both lat and lon are zeroes is another story.

rapotek:
@deg-pl ignored this PR and made a client-side validation, so it won't bother me anymore. The Prime meridian problem (OCUK f.ex.) in newcache still remains, and not only there.

PR #2346, opened 2022-03-18
Fix for reactivation rule option selection problem (fixes #2345)

### LINKED ISSUE #2345: Editing cache description without previously configured reactivation rule leads to issue with sending form
Steps to reproduce
1. Go to cache description edit (for cache that reactivation rule is not set yet)
2. Set reactivation option that you choose
3. Try to send form
4. Sending is not possible because form is still asking for reactivation rule.

Screenshot_64

rapotek:
Replaced "required" "oninvalid" and "oninput" attributes for each reactivation rule radio with ones setting and clearing custom validity only the first input belonging to common radio set.
`cacheDescEditForm`: only a minority of changes is substantive for this fix, the rest are code style fixes.

### DIFF
--- a/src/Views/cacheDescEdit/cacheDescEditForm.tpl.php
+++ b/src/Views/cacheDescEdit/cacheDescEditForm.tpl.php
@@ -1,10 +1,12 @@
 <?php
-use src\Utils\View\View;
+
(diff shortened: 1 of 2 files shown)


deg-pl:
@rapotek thanks for your contribution!

PR #2369, opened 2022-10-19
Prevents geopath multiple publishing, shows correctly status change after button click, fixes #1776

### LINKED ISSUE #1776: Duplicated opening logs on geopaths
From [forum](https://forum.opencaching.pl/viewtopic.php?p=142212#p142212):

> Nie wiem czy temat był już poruszany, ale... może dało by się rozwiązać problem z dublowaniem wpisów przy otwieraniu ścieżek. Zdaję sobie sprawę, że to pierdółka, ale trochę irytująca. Widziałem, że to nie tylko u mnie pojawia się takie cudo. Dzięki.
image

rapotek:
Substantive changes:

- `powerTrail/ajaxUpdateStatus.php` returns correct JSON now instead of HTML,
- `src/Models/PowerTrail/PowerTrail.php`: `canBeOpened()` method additionally checks if geopath is published already

The rest of changes are minor logical fixes (like removing `this` in `static` method) and code style adjustments.

Remark: IMHO these files need a little deeper refactoring, but the whole geopath creation functionality is promised to be rewritten, so let it be as it is for now.

### DIFF
--- a/powerTrail/ajaxUpdateStatus.php
+++ b/powerTrail/ajaxUpdateStatus.php
@@ -1,87 +1,114 @@
 <?php
 
-use src\Utils\Database\OcDb;
-use src\Models\PowerTrail\PowerTrail;
-use src\Utils\Generators\Uuid;
 use src\Models\ApplicationContainer;
+use src\Models\PowerTrail\PowerTrail;
+use src\Utils\Database\OcDb;
+use src\Utils\Generators\Uuid;
 
-require_once __DIR__.'/../lib/common.inc.php';
+require_once __DIR__ . '/../lib/common.inc.php';
+
+$updateStatusResult = [
+    'updateStatusResult' => false,
+    'message' => '',
+];
 
 $loggedUser = ApplicationContainer::GetAuthorizedUser();
-if(!$loggedUser){
-    echo "User not authorized!";
-    exit;
-}
-$ptAPI = new powerTrailBase;
 
-$powerTrailId = (int) $_REQUEST['projectId'];
-$powerTrail = new PowerTrail(array('id' => $powerTrailId));
-$newStatus = (int) $_REQUEST['newStatus'];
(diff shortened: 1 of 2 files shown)


src/Models/PowerTrail/PowerTrail.php
```diff
@@ -2,54 +2,75 @@
...
     private $image;
+
     private $type;
+
     private $centerCoordinates;
+
     private $status;
 
-    /* @var $dateCreated \DateTime */
+    /** @var DateTime */
```
deg-pl, line 44: Why not
`private DateTime $dateCreated;`

rapotek: Probably because this was suggested by cs fixer. I did not analysed most of the code, only tried to pass checks in the end.

powerTrail/ajaxUpdateStatus.php
```diff
@@ -1,87 +1,114 @@
...
+            $updateStatusResult['message'] = tr('pt241');
+        }
+    }
+
+    $updateStatusResult['currentStatus'] = $powerTrail->getStatus();
+    $updateStatusResult['currentStatusTranslation']
+        = $powerTrail->getStatusTranslation();
 } else {
-    print $updateStatusResult['message'];
+    $updateStatusResult['message'] = 'User not authorized!';
```
deg-pl, line 110: You could also add something like this:
http_response_code(HttpCode::STATUS_UNAUTHORIZED);

deg-pl:
Thanks @rapotek !
It is next step to better OC PL code. I think we can merge this PR. What do you think about it @kojoty ?

powerTrail/ajaxUpdateStatus.php
rapotek: You are right, I will try to add it later.

PR #2385, opened 2023-05-20
New news raw view to display news as a separate subpage

kojoty:


src/Models/News/News.php
```diff
@@ -479,9 +479,13 @@ public function getDateLastModified($asString = false)
      *
      * @return string
      */
-    public function getNewsUrl()
+    public function getNewsUrl($rawNews=false)
     {
-        return SimpleRouter::getLink('News.NewsList','show', $this->getId());
+        if ($rawNews) {
```
rapotek, line 484: I know it is merged already, but I would do it like this or similar:
`return SimpleRouter::getLink($rawNews ? 'News.RawNews' : 'News.NewsList', 'show', $this->getId());`

PR #2390, opened 2023-06-13
Parsing uploaded GPX with validation and Topografix, GS 1.0.1 and OC data extraction

rapotek:
`newcacheAjaxWaypointUploader.php` is replaced by API controller, where uploaded file is validated against Topografix XSD 1.0 and 1.1 and the document is processed extracting selected data from Topografix, Groundspeak 1.0.1 and Opencaching, overwriting corresponding element by priority.
An informational popup with parsing rules is added to newcache template.


stefopl:
I’ve checked it and looks good. I’ll merge it today.

It would be nice to add support for http://www.groundspeak.com/cache/1/0
I downloaded a GPX file from a random GC cache, and it used this exact namespace.

It seems like a good idea for the GC code from GS to populate the Foreign waypoints (optional):  Geocaching.com

andrixnet:
> It would be nice to add support for http://www.groundspeak.com/cache/1/0 I downloaded a GPX file from a random GC cache, and it used this exact namespace.
> 
> It seems like a good idea for the GC code from GS to populate the Foreign waypoints (optional): Geocaching.com

Good idea.
IMO it would be better for us (ie opencaching-pl project) to be the more compatible and more inter-operable.

rapotek:
> It would be nice to add support for http://www.groundspeak.com/cache/1/0 I downloaded a GPX file from a random GC cache, and it used this exact namespace.

I could not find schema for 1.0 to provide validation. I don't know if such schema exists at all. As for namespace 1.0 or 1.0.1: in geocaching.com, settings->preferences you can select what version to use for GPX.
 
> It seems like a good idea for the GC code from GS to populate the Foreign waypoints (optional): Geocaching.com

I don't understand, please explain.

stefopl:
@rapotek 

> I could not find schema for 1.0 to provide validation. I don't know if such schema exists at all. As for namespace 1.0 or 1.0.1: in geocaching.com, settings->preferences you can select what version to use for GPX.

I don't use geocaching.com, and this PR was actually the first time I tested creating a cache from a GPX file.
I've had a GC profile for many years, but I don't use it at all and haven't changed any of its settings.

I downloaded a random cache https://www.geocaching.com/geocache/GC952B1 as GPX and got this: https://pastebin.com/raw/vxj1YrBM
The schema seems to be: http://static.groundspeak.com/cache/1/0/cache.xsd

Import worked, but the name was just the GC code, coordinates were fine, and the name appeared as the description.
After changing the code to treat version 1.0 like 1.0.1, the result was much better.

> I don't understand, please explain.

If someone is importing a GPX from GC, I assume they've created a cache there and are now making a related one on OC.
So I think it's useful to extract the GC code from the GPX and fill the field: ```Foreign waypoints (optional): Geocaching.com:```

As I wrote, I don’t use GC and don’t import GPX to OC, so I might have made a mistake.

image

rapotek:
@stefopl Thanks for Groundspeak 1.0 link, I should have been able to find out it by myself. There is only one change between 1.0 and 1.0.1, unimportant for us, so I added 1.0 support including filling up GC waypoint and some fixes [here](https://github.com/opencaching/opencaching-pl/pull/2501)

PR #2410, opened 2024-07-31
Update PHP-CS-Fixer configuration

jrmajor:
- Update PHP-CS-Fixer,
- Only check `src` directory (as mentioned in https://github.com/opencaching/opencaching-pl/pull/2297#issuecomment-940488629)
- Disable all rules except PSR-12 and all rules from PSR-12 that aren't currently satisfied,
- Use PHP-CS-Fixer cache in Style workflow,
- Always check all files in Style workflow, not only the changed ones.

Disabling all rules except for the passing ones leaves us with green CI even if we check all files. Contributors will no longer be forced to reformat every changed file, making diffs more readable. The disabled rules can be fixed and enabled one by one in follow-up PRs.


jrmajor:
Couldn't get the workflow to run on this PR. Maybe it will only run on PRs after it's merged? Seems like it ran in my fork: https://github.com/jrmajor/opencaching-pl/commits/php-cs-fixer-config.

PR #2445, opened 2024-12-05
feature: Add info notices and nearby cache checks to editcache page

stefopl:
Added `info_about_cache_types`, `check_nearby_caches_map`, `check_nearby_caches`, and `check_nearby_caches_info` to the editcache page, aligning its functionality with the newcache page.


andrixnet:
Is this information-only or does this impose restrictions?

stefopl:
The editcache page has been updated to include the same information and buttons "Check for nearby caches..." as the newcache page. 

The only change in functionality is that the Coordinates field is now required on the editcache page, just like on the newcache page.

image

PR #2462, opened 2025-01-08
db charset

stefopl:
@deg-pl What do you think? This should temporarily fix the automatic code updates on OCPL.

### DIFF
--- a/lib/settings-example.inc.php
+++ b/lib/settings-example.inc.php
@@ -57,6 +57,7 @@ $dbserver = 'localhost';
 $dbname = 'ocpl';
 $dbusername = 'ocdbu';
 $dbpasswd = 'PassworD';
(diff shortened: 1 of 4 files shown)


deg-pl:
Great (and simple) idea!

deg-pl:
@stefopl OC PL now is configured to use utf8mb4
post-commit works without any problems.
Thanks!

stefopl:
@deg-pl Thanks

andrixnet:
I believe at least the following fields should receive the same collation: 
- short description
- long description
- waypoint description
- log entry text
- cache personal note
- image name (title)
- geopath title / description
(at least)

Should I go ahead and change them manually, or will there be a DB update issued soon? 

Thank you.

PR #2472, opened 2025-02-19
(fix) Suspected change in IN() behavior affecting GeoPaths sorting

stefopl:


andrixnet:
To what change in IN() behaviour are you referring to ?

stefopl:
@andrixnet 
It seems that when a user has a very large number of found GeoPaths, an IN() query with a list of around 1200 items does not work correctly. Instead of returning the paths in the expected order (based on find dates), they are returned in the default order (by ID).

My test database didn’t contain such a case, so I couldn’t reproduce the issue initially. This PR didn’t solve the problem. Now that I understand more, I will replace the IN() query with a JOIN, as I know it will work correctly. I will also check at what threshold the issue starts occurring. I haven't found any documentation on a limit that could affect this query's behavior.

andrixnet:
I understand. 
Note: it may also be version specific, MariaDB vs MySQL implementation specific or a yet undiscovered bug (?). 
Keep in touch.

PR #2519, opened 2025-05-30
DB update functionality - PSR autoload pattern complaints fix

rapotek:
- Template of class generated by DB updater changed to anonymous class, to fix PSR autoload pattern complaints (class name not corresponding with file name) and keep the functionality of filename change.
- Change filesystem permissions of a newly generated db update file to the ones of the template file. When `template.php` has f.ex. `-rw-rw-r--` permissions and the development user belongs to the group of www-server which generated the file, it makes easier to edit the file.
- Code style changes to delight php-cs-fixer.

### DIFF
--- a/src/Utils/Database/DbUpdates.php
+++ b/src/Utils/Database/DbUpdates.php
@@ -15,15 +15,16 @@ class DbUpdates
 {
     /**
      * @param array|null dict $uuid => DbUpdate object;
-     *                 ordered ascending by filename; read access only via get() / getAll()
+     *                 ordered ascending by filename; read access only via
+     *                 get() / getAll()
      */
     private static ?array $updates = null;
 
     /**
      * @return string
-     *                next unused version number (first part of filename) for a new
-     *                update script to be created; number 900+ is reserved for
-     *                updates that are run always and last
+     *                next unused version number (first part of filename) for a
+     *                new update script to be created; number 900+ is reserved
+     *                for updates that are run always and last
      * @throws Exception
      */
     private static function getNextVersionNumberString(): string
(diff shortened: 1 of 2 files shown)


andrixnet:
Does this require manual intervention by server admins?

rapotek:
> Does this require manual intervention by server admins?

No, these are code changes only, and for future db updates

src/Utils/Database/DbUpdates.php
```diff
@@ -49,18 +50,28 @@ public static function create($developerName): string
...
+        $perms = fileperms($templatePath);
+
+        if ($perms) {
+            $perms = $perms & 0xfff;
+            // Ensure the created file has the same permissions as template.
+            // In particular, when the template has 'rw' for group, and current
+            // code developer belongs to the same group as www server, it makes
+            // easier to modify the created file both by the code developer and
+            // www server.
+            chmod($path, $perms);
```
stefopl, line 73: I have the project directory on Windows and I run the environment using WSL1 (Ubuntu). The project folder is mounted under /mnt/c/..., and in this setup, the chmod() function triggers the error: chmod(): Operation not permitted.
```
set_error_handler(function () {});
chmod($path, $perms);
restore_error_handler();
```

stefopl: It’s been a while since I set up this local environment, but I think on WSL1 I have most things configured like they were before on the VM.
The biggest difference is that I keep the files on Windows for convenience when using the IDE. For the same reason, I also placed dynamic-files directly in the main project folder.

From the Ubuntu side, the drives and folders are automatically mounted. To keep the Apache virtual host structure, I created symlinks in /srv.
image

I chose WSL1 mainly because WSL2 had a less convenient network setup. If I remember correctly, WSL2 used a different IP subnet and was more isolated. On WSL1 everything (SSH, MySQL, Apache2, Xdebug) is accessible on 127.0.0.1, which is easier to work with.

rapotek: I don't know about differences between WSL 1 and 2, because the very first time I met this name was in your post. I had to google out what is it. My dev environment is on native Ubuntu installation, so I did not bothered about alternatives and I was simple curious how this solution works.

I will try to simulate this "permission denied" problem in my system and find out any working solution.

rapotek: > Ok. I’ll check if I changed anything on my local dev and also test it on my VPS dev. Which PHP version are you testing on? I was using PHP 8.3.

8.3.14

stefopl: Sorry for wasting your time. It was my mistake. I copied your changes manually and forgot to add ```use Throwable;```. Without it, PHP doesn’t recognize ```Throwable``` and treats it like an undefined class, so it behaves as if there is no try-catch at all

Instead, I usually use ```catch (\Throwable $ignore)``` and in that case there's no need for the ```use Throwable;```, so I didn't notice the difference earlier.

Issue #2254, opened 2021-04-12
Edit cache - add date picker same as for New cache page


andrixnet:
The "edit cache" page has 2 editable date fields:
- date hidden
- publish on ...

Both fields are for typing manually the date in separate text fields, one for day, one for month, one for year.

The "new cache" page has a single field for a date (YYYY-MM-DD format) and a calendar picker helper. 
Please port this functionality to edit cache page as well.

andrixnet:
![Image](https://github.com/user-attachments/assets/3429f968-1edd-45dc-8028-8b59264c1570)

Date hidden field in "new cache" page. 
- has date picker
- has format information note
- is single form control

![Image](https://github.com/user-attachments/assets/611e31d5-be13-4095-8f4d-bba6742175df)

Date hidden in "edit cache" page.
- does not have date picker
- is split into 3 entry fields
- entry fields are in reverse order to that of the same field in "new cache" 
- entry fields are in reverse order to those of "publish on" fields in "edit cache"
- does not have format information note

---
On "new cache" page, if this field is empty, current date is used. 

Additionally, if field is user set, I suggest a cookie is used to remember this in a similar way to "log entry" and the date_hidden field be autofilled with value stored in cookie . If I want to publish a series of caches it is likely they were hidden on the same day. If that day is not today, it would help me (ie save time with properly pre-filled values).

andrixnet:
For consistency I recommend the same field, the same helper information and the same logic be used on "new cache" and "edit cache" pages.

rborowski:
Hi, just made PR to fix inconsistency with these date inputs. I look forward to a merge/feedback.
I have also taken into account log pages, where date and time was split into 5 fields. I thought that it would be convenient to have date/time pickers in the place we visit most of the time.

stefopl:
There is one issue to fix in the publish_later section on editcache for caches that are not yet published, the datepicker does not work and the input field is too wide. I've marked this with a comment in the PR.

Also, on newcache/editcache, activate_hour is a select dropdown, while on log.php there is a logTimePicker. It might be worth unifying these.

PR #2536, opened 2025-07-20
Add datepickers and unify date (time) inputs

rborowski:
#2254 
### This PR unifies date (time) input fields and adds date (time) pickers across several pages:
- newcache.php
- editcache.php
- log.php
- editlog.php

Instead of having 3 or 5 separate fields for day, month, year, hour and minute, now there is 1 or 2 fields with a picker.

There is also this feature:

> Additionally, if field is user set, I suggest a cookie is used to remember this in a similar way to "log entry" and the date_hidden field be autofilled with value stored in cookie . If I want to publish a series of caches it is likely they were hidden on the same day. If that day is not today, it would help me (ie save time with properly pre-filled values).

I just benchmarked the code from log.php, where user-entered date is stored for a short while in the session.

I have also updated translations accordingly to fit the unified format (YYYY-MMM-DD).

### DIFF
--- a/src/Views/editcache.tpl.php
+++ b/src/Views/editcache.tpl.php
@@ -1,11 +1,27 @@
 <?php
 use src\Utils\View\View;
 use src\Utils\Uri\SimpleRouter;
+
+$view->callChunk('timepicker');
 ?>
 <script>
     $(function () {
         // load country based on coords
         chkcountry2();
+        
+        // Initialize timepicker
+        $('#activateTimePicker').timepicker({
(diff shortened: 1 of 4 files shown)


src/Views/editcache.tpl.php
```diff
@@ -274,7 +211,7 @@ function selectPublishLater(){
 </script>
 
 <style>
-  #hiddenDatePicker, #activateDatePicker{
```
stefopl, line 277: This should remain unchanged.

